PR #33, opened 2015-03-17
Updated setup.py to install README.md instead of README

jkutemeier:
setup.py fails because there no longer is a README file.


jmcgeheeiv:
Thank you for the pull request.  I already merged an identical one.

Sorry for the late response.  It turns out that I was not watching my own project!

PR #34, opened 2015-03-19
Fix setup.py build failure

orsenthil:
Without this change, the setup.py build will result in

  File "setup.py", line 19, in <module>
    LONG_DESCRIPTION = open(readme).read()
IOError: [Errno 2] No such file or directory: 'README'


jmcgeheeiv:
Thank you for the pull request.  I merged it.

Sorry for the late response.  It turns out that I was not watching my own project!

PR #39, opened 2015-04-18
Normalize path before removing files. Fixes #31

### LINKED ISSUE #31: os.remove doesn't work with relative paths
```
The 'remove' method in the FakeOsModule doesn't support relative paths.

What steps will reproduce the problem?

fake_os.mkdir("foo")
fake_os.mkdir("foo/bar")  # works
fake_os.chdir("foo")  # works
fake_os.remove("bar") # remove looks for "bar" at the root

What is the expected output? What do you see instead?
expected: file is deleted
actual outcome: 'bar' does not exist in fake filesystem

What version of the product are you using? On what operating system?
2.4, Windows 7

Please provide any additional information below.

```

Original issue reported on code.google.com by `londinop` on 25 Nov 2014 at 3:28

megapctr:


megapctr:
Could you retrigger the build? It failed, but not because of my changes.

jmcgeheeiv:
I see that the build has been green for about three days.  Where are you seeing this failure?

jmcgeheeiv:
Ah, now I see.  Until recently pyfakefs did not with versions other than Python 2.7.  You can probably get your branch to work by merging in my changes:

    git pull origin master

jmcgeheeiv:
Thanks for the fix, Filip. It's great to see the bug backlog going down.

PR #40, opened 2015-04-22
Treat cygwin as a Windows

megapctr:
To be more specific, I had errors with `os.path.realpath`. As it turned out, cygwin wasn't recognized as a Windows and thas was causing the errors.

This commit should fix most of other undiscovered issues with cygwin.
Also, it's better to keep the critical logic of recognizing the system in a single place.

### DIFF
--- a/fake_filesystem.py
+++ b/fake_filesystem.py
@@ -126,9 +126,10 @@ FAKE_PATH_MODULE_DEPRECATION = ('Do not instantiate a FakePathModule directly; '
 class Error(Exception):
   pass
 
+is_windows = sys.platform.startswith('win') or sys.platform == 'cygwin'
 
-if sys.platform.startswith('win'):
-  # On Windows, raise WindowsError instead of OSError
+if is_windows and 'WindowsError' in globals():
+  # On Windows, raise WindowsError instead of OSError if available
   OSError = WindowsError  # pylint: disable-msg=E0602,W0622
 
 
@@ -1118,7 +1119,7 @@ class FakePathModule(object):
     """Normalize path, eliminating double slashes, etc."""
     return self.filesystem.CollapsePath(path)
 
-  if sys.platform.startswith('win'):
+  if is_windows:
 
     def relpath(self, path, start=None):
       """ntpath.relpath() needs the cwd passed in the start argument."""
--- a/fake_filesystem.py
+++ b/fake_filesystem.py
@@ -126,9 +126,10 @@ FAKE_PATH_MODULE_DEPRECATION = ('Do not instantiate a FakePathModule directly; '
 class Error(Exception):
   pass
 
-is_windows = sys.platform.startswith('win') or sys.platform == 'cygwin'
+_is_windows = sys.platform.startswith('win')
+_is_cygwin = sys.platform == 'cygwin'
 
-if is_windows and 'WindowsError' in globals():
+if _is_windows:
   # On Windows, raise WindowsError instead of OSError if available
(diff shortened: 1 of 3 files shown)


fake_filesystem.py
```diff
@@ -126,9 +126,10 @@
 class Error(Exception):
   pass
 
+is_windows = sys.platform.startswith('win') or sys.platform == 'cygwin'
 
-if sys.platform.startswith('win'):
-  # On Windows, raise WindowsError instead of OSError
+if is_windows and 'WindowsError' in globals():
```
megapctr: WindowsError is not available in cygwin

jmcgeheeiv:
Within fake_filesystem.py, how about makint it a private variable _is_windows?

Of course that means that the definition would have to be repeated in fake_filesystem_test.py, but I think that is acceptable in a test file.  How about adding is_windows to class fake_filesystem_test.TestCase?

Also, could you verify that all those tests really must be skipped in Cygwin?  Do some of the tests skipped for Windows actually pass on Cygwin?  I think you already know the answer to this.

Thanks for your help on this.

megapctr:
All tests pass, with the exception of testExpandUser (the "~root" bit fails).

After some tinkering with my original problem, it turned out that the issue with `os.path.realpath` is even bigger, and possibly not related to cygwin

The exception I'm getting is `AttributeError: 'NoneType' object has no attribute 'error'`

Pytest traceback is useless, even though I've added `py` and `inspect` modules to `Stubber.SKIPMODULES` set.
```
/home/Filip/renamer/renamer.py:30: in common_dir
    paths = [os.path.realpath(p) for p in paths]
/home/Filip/pyfakefs/fake_filesystem.py:1142: in foo
    ???
/home/Filip/.virtualenvs/renamer/lib/python2.7/posixpath.py:382: in realpath
    ???
/home/Filip/.virtualenvs/renamer/lib/python2.7/posixpath.py:407: in _joinrealpath
    ???
```

I've modified `FakePathModule.__getattr__` function to wrap result in a exception-catching, traceback-printing decorator:
```
   def __getattr__(self, name):
     """Forwards any non-faked calls to os.path."""
     print name
     attr = self._os_path.__dict__[name]
     def foo(*a):
         print name, a
         try:
             result = attr(*a)
         except Exception as e:
             print "TRACEBACK"
             formatted_lines = traceback.format_exc().splitlines()
             print '\n'.join(formatted_lines)
             raise
         return result
     return foo
```

Here's the traceback:
```
realpath (u'bar',)
TRACEBACK
  File "/home/Filip/pyfakefs/fake_filesystem.py", line 1142, in foo
  File "/home/Filip/.virtualenvs/renamer/lib/python2.7/posixpath.py", line 382, in realpath
  File "/home/Filip/.virtualenvs/renamer/lib/python2.7/posixpath.py", line 407, in _joinrealpath
  File "/home/Filip/.virtualenvs/renamer/lib/python2.7/posixpath.py", line 143, in islink
AttributeError: 'NoneType' object has no attribute 'error'
```

Here's standard lib's posixpath.islink code:
```
def islink(path):
    """Test whether a path is a symbolic link"""
    try:
        st = os.lstat(path)
    except (os.error, AttributeError):
        return False
    return stat.S_ISLNK(st.st_mode)
```

I've then checked and confirmed that in this scope `os` is actually a `None`. At this point I'm totally clueless. What could be causing this?

megapctr:
I've fixed that issue and created a separate PR. See #41

megapctr:
I've added the changes you requested. Is everything okay now?

jmcgeheeiv:
Nice job.  Pushed.

PR #41, opened 2015-04-25
Fix FakePathModule deprecation in Stubber

megapctr:
`Stubber` class was instantiating `FakePathModule` directly, without passing `os_module` argument. FakePathModule docstring states that such behavior is deprecated.


jmcgeheeiv:
And now you fix my code!  Thanks.

PR #51, opened 2015-09-06
Typo in AUTHOR_EMAIL

tomviner:
(If this was a deliberate anti spam strategy, maybe add a comment)


jmcgeheeiv:
What an excellent idea for an anti-spam strategy.  No, just a typo.  Thanks.

Issue #54, opened 2015-11-11
AttributeError: 'TestClass' object has no attribute 'addCleanup'


petergillardmoss:
This is because unittest2 is not imported on python < 2.7.

PR #55, opened 2015-11-11
Use unittest2 on python < 2.7

petergillardmoss:
Fixes https://github.com/jmcgeheeiv/pyfakefs/issues/54


jamescooke:
Looks good to me :+1:

Issue #56, opened 2015-12-06
travis build reports green when there are failures


odahoda:
See e.g. https://travis-ci.org/jmcgeheeiv/pyfakefs/jobs/90503441:

    [...lots of output...]
    ======================================================================
    ERROR: testByteContents (fake_filesystem_test.FakeFileOpenTest)
    ----------------------------------------------------------------------
    Traceback (most recent call last):
      File "/home/travis/build/jmcgeheeiv/pyfakefs/fake_filesystem_test.py", line 2079, in testByteContents
        f.write(byte_fractions)
    TypeError: string argument expected, got 'bytes'
    ----------------------------------------------------------------------
    Ran 402 tests in 0.265s
    FAILED (errors=1, skipped=3)
    The command "./all_tests.py" exited with 0.
    Done. Your build exited with 0.

all_tests.py does not return a non-zero return code when there were errors in the test suite.
TestRunner.run() returns a TestResult object and you have to return 1 from the script and not result.wasSuccessful().

jmcgeheeiv:
Nice catch.  Note that this uncovered a problem with byte strings as described in issue #63.

PR #60, opened 2016-01-08
Protect binary data a little better

dougluce:
This also works for non-Unicode encodings (like compressed data).  

It fixes an expected behavior on Python 3 where reading strings from a binary file opened in non-binary mode will raise an exception.

### DIFF
--- a/fake_filesystem.py
+++ b/fake_filesystem.py
@@ -89,6 +89,8 @@ import stat
 import sys
 import time
 import warnings
+import binascii
+
 try:
   import cStringIO as io  # pylint: disable-msg=C6204
 except ImportError:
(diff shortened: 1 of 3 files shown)


jmcgeheeiv:
This is just what we need.

However, due to #56, *Travis build reports green when there are failures*, the tests are not actually green.  If you look at the end of the detailed report, you can see errors.  By no means are all the failures yours.  I cleaned up some.

Please pull the latest pyfakefs again.  After you do this, the errors that I fixed will disappear, and Travis will fail if there are errors.

jmcgeheeiv:
Doug, might you consider working on #63?  You seem to be qualified.

dougluce:
I'll take a look at the failures after rebasing in a bit.  And see if #63 is easy to knock off.

PR #61, opened 2016-01-09
Add TemporaryDirectory to the tempfile module

dougluce:
Also add a test to make sure the unittest module picks it up.

### DIFF
--- a/fake_filesystem.py
+++ b/fake_filesystem.py
@@ -149,6 +149,20 @@ def CopyModule(old):
   return new
 
 
+class Hexlified(object):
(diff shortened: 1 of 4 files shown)


jmcgeheeiv:
This is an awesome enhancement.

However, due to #56, *Travis build reports green when there are failures*, the tests are not actually green.  If you look at the end of the detailed report, you can see errors.  By no means are all the failures yours.  I cleaned up some.

Please pull the latest pyfakefs again.  After you do this, the errors that I fixed will disappear, and Travis will fail if there are errors.

Issue #64, opened 2016-01-13
pyfakefs not installing from pypi


petergillardmoss:
Hello,

I get the following error when trying to install pyfakefs from pypi:

```
$ pip install pyfakefs
Collecting pyfakefs
  Downloading pyfakefs-2.6.tar.gz
    Complete output from command python setup.py egg_info:
    Traceback (most recent call last):
      File "<string>", line 20, in <module>
      File "/private/var/folders/59/y54yv6md5h36hkw9pc537rf80000gn/T/pip-build-wMslO8/pyfakefs/setup.py", line 37, in <module>
        LONG_DESCRIPTION = open(readme).read()
    IOError: [Errno 2] No such file or directory: '/private/var/folders/59/y54yv6md5h36hkw9pc537rf80000gn/T/pip-build-wMslO8/pyfakefs/README.md'
```

jcwilson:
See #45

jmcgeheeiv:
Please try pyfakefs 2.7.

jmcgeheeiv:
@jcwilson, @cburroughs, and Cash Costello confirmed that pyfakefs 2.7 on PyPi works.

Thank you for your patience.

Issue #45, opened 2015-05-11
The pyfakefs on PyPi is obsolete 


spg:
So I `pip install pyfakefs`
Then I `import fake_filesystem_unittest` in one of my test modules
And I get `ImportError: No module named fake_filesystem_unittest` when running tests with `py.test`

Is that expected?

megapctr:
The pyfakefs on PyPy is old. Try installing pyfakefs through github url, this should help

centralniak:
Perhaps it would be better to release a set of unittest pyfakefs tools as a separate package so that there is no need to overwrite original `pyfakefs` on PyPI?

jmcgeheeiv:
in [davidlbaird/pyfakefs issue #33](https://github.com/davidlbaird/pyfakefs/issues/33), I have asked Mr. Baird to allow me access to PyPi pyfakefs.  If you think this is a good idea, please add your opinion as a comment to [davidlbaird/pyfakefs issue #33](https://github.com/davidlbaird/pyfakefs/issues/33).

jmcgeheeiv:
I am considering releasing pyfakefs as pyfakefs3, a name I control.  Comments?

jamescooke:
:+1:  for `pyfakefs3`. I hope that releasing the new version on PyPI will help move the project forwards.

holtgrewe:
:+1: 

Please upload to pypi, I cannot deploy it organisation-wide otherwise

cash:
PyPI lists jmcgeheeiv as the maintainer so @jmcgeheeiv should be able to update the package. https://pypi.python.org/pypi/pyfakefs

cburroughs:
Apologizes for the bump, but I'm using a non-pip system where going around PyPi is extra difficult (pants).  I think pyfakefs3 is fine, if the PyPi metadata means @jmcgeheeiv  has access to the existing package after all, so much the better.

jmcgeheeiv:
I apologize for the delay.  I will get to this soon.  I see that I have quite a corps of testers to make sure PyPi works smoothly.

jmcgeheeiv:
Okay, ladies and gentlemen, give [pyfakefs on PyPi](https://pypi.python.org/pypi/pyfakefs) a try and reopen this issue if you find problems.

cburroughs:
[...]

PR #65, opened 2016-01-13
Add a pyfakefs/ module directory; include README.md and COPYING files

jcwilson:
This should enable PyPi usage.

### DIFF
--- a/MANIFEST.in
+++ b/MANIFEST.in
@@ -1,0 +1,2 @@
+include COPYING
+include README.md
--- a/README.md
+++ b/README.md
@@ -11,7 +11,7 @@ for a concrete example of how to apply pyfakefs.
 Usually you would use it this way:
 
 ```python
-import fake_filesystem as fake_fs
+import pyfakefs.fake_filesystem as fake_fs
 
 # Create a faked file system
 fs = fake_fs.FakeFilesystem()
@@ -28,7 +28,7 @@ try:
 except ImportError:
     from mock import patch  # Python 2
 
-import fake_filesystem_glob as fake_glob
+import pyfakefs.fake_filesystem_glob as fake_glob
 
 # Note that this fake module is based on the fake fs you just created
 glob = fake_glob.FakeGlobModule(fs)
@@ -38,7 +38,7 @@ with patch('mymodule.glob', glob):
 
 ## Continuous Integration
 
-pyfakefs is automatically tested with Python 2.6 and above, and it is currently 
+pyfakefs is automatically tested with Python 2.6 and above, and it is currently
 [![Build Status](https://travis-ci.org/jmcgeheeiv/pyfakefs.svg)](https://travis-ci.org/jmcgeheeiv/pyfakefs).
 
 See [Travis-CI](http://travis-ci.org) for
--- a/example.py
+++ b/example.py
@@ -20,19 +20,19 @@ This demonstrates the usage of the
 
 The modules related to file handling are bound to the respective fake modules:
 
->>> os     #doctest: +ELLIPSIS 
-<fake_filesystem.FakeOsModule object...>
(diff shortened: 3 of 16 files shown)


jcwilson:
For #45 and #64

jcwilson:
I was able to test this by doing:
```
$ python setup.py sdist
$ pip install file:///shared/pyfakefs/dist/pyfakefs-2.6.tar.gz#egg=pyfakefs-2.6
```

jmcgeheeiv:
I have been looking into this.  I created a MANIFEST.in:

more MANIFEST.in
include README.md
include COPYING
include __init__.py

Do note that in the current uploaded version, __init__.py is missing.
However, it is my understanding that putting a file in MANIFEST.in does not
get it installed.  I have to figure out that as well.

jcwilson:
added those in, too

jcwilson:
interesting, i guess the docs are a little misleading about having to explicitly name the rest of your files.  Seems like your slimmed-down one suffices. Let's go with that

jmcgeheeiv:
Did you install this and observe that `__init__.py` is *installed*?

jcwilson:
the way this is being installed is that each of the `fake_*` files are being copied directly into site-packages (along with pyfakefs-2.6-py2.7.egg-info being created there). The `__init__.py` is irrelevant in this distribution manner. If we wanted to treat it as an actual package, we'd need to have another directory (named `pyfakefs`) in the repo and move the `fake_*` files into there. I can take a stab at that, but it will be an interface-breaking change for any current users that attempt a simple upgrade.

the usage would become something like:
```
from pyfakefs import fake_filesystem_unittest
```

jmcgeheeiv:
Yes, the absence of a pyfakefs/ directory causes pain in several ways.  I think it's worth breaking existing installations to add the directory.

jcwilson:
most of the tests pass. i think I can chalk the failures up to my running them from cygwin and using nosetests. I'm not sure what test runner you're using.

jmcgeheeiv:
Travis and I run all_tests.py.  I have heard reports of weirdness with Cygwin.

Anyway, if you got most of the tests to run and pass, that is a fine test of the installation.  I'll upload pyfakefs again.

jmcgeheeiv:
Very nice.  Thanks for taking care of that.

Could you please fix this test failure?

$ ./all_tests.py
  File "./all_tests.py", line 21, in <module>
    import fake_filesystem_glob_test
  File "/home/travis/build/jmcgeheeiv/pyfakefs/fake_filesystem_glob_test.py", line 27, in <module>
    from pyfakefs import fake_filesystem_glob
  File "/home/travis/build/jmcgeheeiv/pyfakefs/pyfakefs/fake_filesystem_glob.py", line 38, in <module>
    import fake_filesystem
ImportError: No module named fake_filesystem
The command "./all_tests.py" exited with 1.
Done. Your build exited with 1.

jcwilson:
i think i fixed that, but there's something going on with the doc tests. Unfortunately, i don't have python 3 handy, and i'm not familiar enough with the doctest functionality to say why it's failing now.

jcwilson:
want me to collapse these into one commit?

jmcgeheeiv:
If it's not too much trouble.  Let me know and I'll pull your fork and deal
with the doctest/Python3 issue.

jcwilson:
no trouble at all. seems like all the tests are passing now, too :)

also, if you think it's appropriate, you can add `exclude pyfakefs/example.py` to the manifest file to ensure it doesn't ship when installing

jcwilson:
I relocated the `example.py` file back to the top-level directory, since it's not something that should really be considered part of the package

jmcgeheeiv:
Nice work, @jcwilson.  I uploaded this as pyfakefs 2.7.  Please give it a try.

Issue #23, opened 2015-03-16
fake_tempfile is using AddOpenFile incorrectly.


GoogleCodeExporter:
```
Code to reproduce:

def Repro():
  fd, fn = tempfile.mkstemp()
  fo = os.fdopen(fd, 'w+b')
  ...

I'd get a stack trace like this:

"myfile.py", line 3, in Repro
    fo = os.fdopen(fd, 'w+b')
  File ".../fakefs/fake_filesystem.py", line 1208, in _fdopen_ver2
    return FakeFileOpen(self.filesystem).Call(file_des, mode=mode)
  File ".../fakefs/fake_filesystem.py", line 1918, in Call
    file_object = self.filesystem.GetOpenFile(filedes).GetObject()
AttributeError: 'FakeFile' object has no attribute 'GetObject'

The offending part in fake_tempfile calls AddOpenFile() with a FakeFile object, 
but according to fake_filesystem.py line 2184 this should be FakeFileWrapper 
which is only visible from the FakeFileOpen class.

liulk
```

Original issue reported on code.google.com by `liulk@google.com` on 6 Dec 2013 at 12:09

jmcgeheeiv:
Fixed by @dougluce in pull request #66.

PR #66, opened 2016-01-14
Wrap file created by mkstemp

dougluce:
This promotes the FakeFileWrapper class to top-level and has mkstemp wrap its file.  This lets fdopen() work properly.  This should address #23 (among other things).

### DIFF
--- a/fake_tempfile_test.py
+++ b/fake_tempfile_test.py
@@ -64,6 +64,7 @@ class FakeTempfileModuleTest(unittest.TestCase):
 
   def setUp(self):
     self.filesystem = fake_filesystem.FakeFilesystem(path_separator='/')
+    self.os = fake_filesystem.FakeOsModule(self.filesystem)
     self.tempfile = fake_tempfile.FakeTempfileModule(self.filesystem)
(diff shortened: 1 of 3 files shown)


jmcgeheeiv:
Thank you for this fix.  Could I ask you to address these?
* Can `class FakeFileWrapper` be returned to its original place in the file so that the diff is smaller?
* Please add a test.  #23 provides a fine start.

dougluce:
I don't think `FakeFileWrapper` can be returned to its original place as it was promoted and unnested from `FakeFileOpen`.  If it helps, this is the diff of the other changes to that class:

```
12c12
<   def __init__(self, file_object, update=False, read=False, append=False,
---
>   def __init__(self, file_object, file_path, update=False, read=False, append=False,
15a16
>     self._file_path = file_path
196c197
<       raise FakeLargeFileIoException(file_path)
---
>       raise FakeLargeFileIoException(self._file_path)
```

I extended the `testMkstemp` test in `fake_tempfile_test.py` to cover this case, adapted from the description on #23.  This test fails on 37fb868 (as is proper).

Issue #24, opened 2015-03-16
Reading from fake block devices doesn't work


GoogleCodeExporter:
```
What steps will reproduce the problem?
1. Create a fake block device (dev = fs.CreateFile('name', stat.S_IFBLK))
2. Try and read from it (dev.read())


What is the expected output? What do you see instead?
This should work, but an error is raised about trying to read from a directory.

Patch with fix and test attached.

```

Original issue reported on code.google.com by `dmwh...@google.com` on 12 Dec 2013 at 3:13

Attachments:
* [block_device_read.patch](https://storage.googleapis.com/google-code-attachments/pyfakefs/issue-24/comment-0/block_device_read.patch)

jmcgeheeiv:
Fixed by @dougluce in pull request #67.

PR #67, opened 2016-01-14
Use proper checking for directory on open.

dougluce:
So block devices aren't accidentally considered directories.  Should address #24.


jmcgeheeiv:
Please add an assertRaises() to test the error when the file is a directory.

dougluce:
Is that path [not already covered?](https://github.com/jmcgeheeiv/pyfakefs/blob/master/fake_filesystem_test.py#L2177-L2180)

PR #68, opened 2016-01-22
Writes update mtime and float time support

humm:
mtime and ctime are updated during the SetContents() methods.
Removes a change to ctime that occurred just before that.

FakeFilesystem gets an `float_stat` option, that updates stat time
to the full precision of the time.time() call. The existing behavior,
using integers, is preserved as the default.

### DIFF
--- a/fake_filesystem_test.py
+++ b/fake_filesystem_test.py
@@ -1400,10 +1400,11 @@ class FakeOsModuleTest(TestCase):
 
   def testUtimeSetsCurrentTimeIfArgsIsNoneWithFloats(self):
     # set up
-    # time.time can report back floats, but it should be converted to ints
-    # since atime/ctime/mtime are all defined as seconds since epoch.
+    # we set os.stat_float_times() to False, so atime/ctime/mtime
+    # are converted as ints (seconds since epoch)
     time.time = _GetDummyTime(200.0123, 20)
     path = '/some_file'
+    fake_filesystem.FakeOsModule.stat_float_times(False)
     self._CreateTestFile(path)
     st = self.os.stat(path)
     # 200 is the current time established above (if converted to int).
@@ -1416,8 +1417,10 @@ class FakeOsModuleTest(TestCase):
     self.assertEqual(240, st.st_mtime)
 
   def testUtimeSetsCurrentTimeIfArgsIsNoneWithFloatsNSec(self):
-    self.filesystem = fake_filesystem.FakeFilesystem(path_separator='/', nsec_stat=True)
+    self.filesystem = fake_filesystem.FakeFilesystem(path_separator='/')
     self.os = fake_filesystem.FakeOsModule(self.filesystem)
+    self.os.stat_float_times(True)
+    self.assertTrue(self.os.stat_float_times())
 
     time.time = _GetDummyTime(200.0123, 20)
     path = '/some_file'
--- a/fake_filesystem_test.py
+++ b/fake_filesystem_test.py
@@ -1415,6 +1415,21 @@ class FakeOsModuleTest(TestCase):
     self.assertEqual(220, st.st_atime)
     self.assertEqual(240, st.st_mtime)
 
+  def testUtimeSetsCurrentTimeIfArgsIsNoneWithFloatsNSec(self):
+    self.filesystem = fake_filesystem.FakeFilesystem(path_separator='/', nsec_stat=True)
+    self.os = fake_filesystem.FakeOsModule(self.filesystem)
+
+    time.time = _GetDummyTime(200.0123, 20)
+    path = '/some_file'
+    self._CreateTestFile(path)
+    st = self.os.stat(path)
+    # 200.0123 not converted to int
+    self.assertEqual(200.0123, st.st_atime, st.st_mtime)
+    self.os.utime(path, None)
+    st = self.os.stat(path)
+    self.assertEqual(220.0123, st.st_atime)
+    self.assertEqual(240.0123, st.st_mtime)
+
   def testUtimeSetsSpecifiedTime(self):
     # set up
     path = '/some_file'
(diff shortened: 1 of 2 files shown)


humm:
This is intended to allow simulating filesystems that have integer- or subsecond-stat support, without messing with time.

jmcgeheeiv:
Thank you for the pull request.

It seems to me that instead of a new argument, [os.stat_float_times()](https://docs.python.org/3/library/os.html#os.stat_float_times) is the place to query/set/reset the type that is returned.  If you add this, we get new standard functionality and we don't have to even document it.

Could you please check this out?

humm:
Yes, that is better. I'll modify the PR accordingly (may take a week or two).

Should float be the default? And should we store float values anyway internally, and return int or float values depending on `os.stat_float_times()`(e.g. via properties)? In other word, should a call in a float filesystem to `os.stat_float_times(False)` and then to `os.stat_float_times(True)` loose the non-integer part of the timestamps or not? My inclination would be to avoid loosing info there.

What do you think?

jmcgeheeiv:
An experiment with Python 3 shows that is starts out in float mode:
```
Python 3.4.3 (default, Oct 14 2015, 20:28:29) 
[GCC 4.8.4] on linux
Type "help", "copyright", "credits" or "license" for more information.
>>> import os
>>> os.stat_float_times()
True
>>> os.path.getatime('/home/jmcgehee/.bashrc')
1459545702.6583292
>>> os.path.getctime('/home/jmcgehee/.bashrc')
1456261286.569785
>>> os.path.getmtime('/home/jmcgehee/.bashrc')
1456261286.565785
>>> os.stat_float_times(False)
>>> os.path.getatime('/home/jmcgehee/.bashrc')
1459545702
>>> os.path.getctime('/home/jmcgehee/.bashrc')
1456261286
>>> os.path.getmtime('/home/jmcgehee/.bashrc')
1456261286
>>> os.stat_float_times()
False
```
With regard to losing information, I believe that you can keep the time as a float, and round down to get the integer value.

jmcgeheeiv:
To be clear about my status, I am waiting on @humm.  But @humm said it would take a week or two, so that's fine.

humm:
So I looked into it, and this seems trickier than I thought. If we are to implement `os.stat_float_times()` into `FakeOsModule`, then we need `FakeFile` instances to call `os.stat_float_times()` whenever the user wants to access `st_mtime`, `st_atime` or `st_ctime`. But currently, `FakeFile` objects are simple, and do not have a reference to `FakeOsModule`. And given the current architecture of `pyfakefs`, it is possible to instanciate different `FakeOsModule` for the same filesystem, leading to potential conflicts over the `os.stat_float_times()` setting.

There are solutions, but they either require to fix the `stat_float_times` setting at a `FakeFile` creation, potentially leading to inconsistency if other `FakeFiles` are created with a different setting (a file created after another could have a earlier creation date because one has its time rounded down and not the other), or to significantly alter the current architecture of `pyfakefs`. The most reasonable way may be to currently set the `stat_float_times` setting only at the creation of Filesystem, but since FakeOsModule is created after, we cannot call `os.stat_float_times()` at that point. And if we rely on the real `os` module, it seems unintuitive, intrusive and potentially problematic.

What is your take on all of this? Maybe I missed an obvious solution?

jmcgeheeiv:
Perhaps if you added a private module attribute `os._stat_float_times`.   This makes it a singleton so it will always have a single global value.  [In Python 2.5, the times were switched to floats](https://docs.python.org/2/library/os.html#os.stat_float_times) so you can initialize it:

    os._stat_float_times = sys.version_info >= (2, 5)

Then: 

-  Files always keep their times as a float
- `os.stat_float_times()` gets and sets the value of `os._stat_float_times`
-  `os.path.getatime()`, etc. return an integer when `os._is_float_times` is `True`

jmcgeheeiv:
Also, the master has moved ahead quite a bit, so if you could merge the current master into your pull request branch, that would help a lot:

    git pull git@github.com:jmcgeheeiv/pyfakefs.git master

humm:
Yes, the class variable is the obvious solution. Of course, I will update my PR to the latest master. I can even rebase it if you prefer.

jmcgeheeiv:
Yes, please help me make it easy to merge.

humm:
Ok, done. Let's note that while we have `_stat_float_times = sys.version_info >= (2, 5)`, pyfakefs is not compatible with Python earlier than 2.6.

jmcgeheeiv:
Indeed I almost recommended `_stat_float_times = True`, but is was so easy to not break this for Python  2.4, I decided to recommend testing for the version.

humm:
I used, however, property setter decorators in this PR that were not introduced until 2.6. It could be modified to work without them, though, but the code would be less clean (setting underscored variables from outside the `FakeFile` class).

fake_filesystem_test.py
```diff
@@ -1400,10 +1400,11 @@ def testUtimeSetsCurrentTimeIfArgsIsNone(self):
 
   def testUtimeSetsCurrentTimeIfArgsIsNoneWithFloats(self):
     # set up
-    # time.time can report back floats, but it should be converted to ints
-    # since atime/ctime/mtime are all defined as seconds since epoch.
+    # we set os.stat_float_times() to False, so atime/ctime/mtime
+    # are converted as ints (seconds since epoch)
     time.time = _GetDummyTime(200.0123, 20)
```
jmcgeheeiv: Use 200.9123 to test that the float is rounded down.

fake_filesystem_test.py
```diff
@@ -1415,6 +1416,23 @@ def testUtimeSetsCurrentTimeIfArgsIsNoneWithFloats(self):
     self.assertEqual(220, st.st_atime)
```
jmcgeheeiv: Also test the type: self.assertTrue(isinstance(st.st_atime, int))`

fake_filesystem_test.py
```diff
@@ -1400,10 +1400,11 @@ def testUtimeSetsCurrentTimeIfArgsIsNone(self):
...
   def testUtimeSetsCurrentTimeIfArgsIsNoneWithFloats(self):
     # set up
-    # time.time can report back floats, but it should be converted to ints
-    # since atime/ctime/mtime are all defined as seconds since epoch.
+    # we set os.stat_float_times() to False, so atime/ctime/mtime
+    # are converted as ints (seconds since epoch)
     time.time = _GetDummyTime(200.0123, 20)
     path = '/some_file'
+    fake_filesystem.FakeOsModule.stat_float_times(False)
     self._CreateTestFile(path)
```
jmcgeheeiv: Please investigate and test the proper behavior when you create a file first, and then set `stat_float_times(False)`

humm:
Ok, so I improved the test, and caught a bug. At file creation, `mtime` and `atime` are copied from `ctime`, and so far were copied from the possibly integer value. They are now copied directly from the internal, full-precision value.

Let's note here that we don't exactly reproduce the behaviour of a HFS+ filesystem for instance, where time values are in float by default (because Python >= 2.5), but because the filesystem only support integer precision, the decimal part is zero. But I guess that's close enough for any practical purposes.

PR #76, opened 2016-03-28
Added support for alternative path separator

### LINKED ISSUE #28: should override os.path.normcase on Windows
```
FakeFilesystem takes a path_separator arg, primarily useful for cross-platform 
faking on Windows.  However it doesn't override os.path.normcase, which on 
Windows replaces forward-slashes with backslashes.

What steps will reproduce the problem?
import fake_filesystem
filesystem = fake_filesystem.FakeFilesystem(path_separator='/')
fake_os = fake_filesystem.FakeOsModule(filesystem)
os = fake_os                    # bash real os
print os.path.normcase('/tmp')

What is the expected output?
/tmp

What do you see instead?
\tmp

What version of the product are you using? On what operating system?
2.4, on Windows

This causes problems with e.g. the tempfile module, because it uses normcase 
which produces invalid filenames in this situation.

I think the simplest solution is to just override os.path.normcase with a null 
implementation that returns its arg on Windows, if the filesystem's 
path_separator is '/'.
```

Original issue reported on code.google.com by `ga...@oberbrunner.com` on 2 May 2014 at 3:41

mrbean-bremen:
- added support for optional additional path separator to allow mixed
separators as under real fs under Windows
- implemented normcase to use the defined path separator
Shall fix #28

### DIFF
--- a/fake_filesystem_shutil_test.py
+++ b/fake_filesystem_shutil_test.py
@@ -43,6 +43,30 @@ class FakeShutilModuleTest(unittest.TestCase):
     self.shutil.rmtree(directory)
     self.assertFalse(self.filesystem.Exists(directory))
 
+  def testRmtreeNonExistingDir(self):
+    directory = 'nonexisting'
+    self.assertRaises(IOError, self.shutil.rmtree, directory)
+    try:
+      self.shutil.rmtree(directory, ignore_errors=True)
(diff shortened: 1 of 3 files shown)


mrbean-bremen:
Mixup of 2 branches, closing.

PR #77, opened 2016-03-30
Added support for additional path separator

mrbean-bremen:


jmcgeheeiv:
Thank you for this pull request.  This looks very valuable.

Is there any situation on the Windows platform that you would not want `alt_path_sep='\\'`?  If not, it seems best to set it automatically.  What do you think?

Also, the variable name uses _alternate_ and the docstrings use _additional_.  Let's use one or the other.

mrbean-bremen:
Yes, I wasn't sure about this. If you are using the default path separator it makes sense to set it automatically, otherwise not. Maybe we should indeed set it automatically in the default case and not   pass it as an argument at all. What do you think? 
About the name - yes, I changed my mind halfway through the code, I will fix this.

pyfakefs/fake_filesystem.py
```diff
@@ -362,8 +362,14 @@ def __init__(self, path_separator=os.path.sep):
 
     Args:
       path_separator:  optional substitute for os.path.sep
+
+      Example usage to emulate real file systems:
+         filesystem = FakeFilesystem(alt_path_separator='/' if _is_windows else None)
     """
     self.path_separator = path_separator
+    self.additional_path_separator = None
+    if _is_windows and path_separator == os.sep:
```
jmcgeheeiv: `path_separator == os.sep`-- very good.

jmcgeheeiv:
This is great work.  After further investigation, I found that module os has [os.altsep](https://docs.python.org/3/library/os.html#os.altsep), which is the same as the `alt_path_separator` that you added.  Note that `os.altsep` is also exposed as `os.path.altsep`.  I noted that on Linux these are both initialized to `None`, which makes sense.

Could you please implement `os.altsep` and `os.path.altsep` and add a test for them?

mrbean-bremen:
Ah, good find! I wasn't aware of os.altsep (though may have seen it). Done.

Issue #83, opened 2016-04-11
Support for drive letter in Windows incomplete


mrbean-bremen:
Under Windows drive letters are not completely supported:
```
>>> filesystem = fake_filesystem.FakeFilesystem(path_separator='/')
>>> filesystem.NormalizePath('c:/foo')
'/c:/foo'
```
I will try to add a fix for NormalizePath - as far as I can see, this should fix most problems. 
What do you think?
(Edit: removed SplitPath() from description which behaves correctly)

mrbean-bremen:
PR #84 fixes this issue for me.

mrbean-bremen:
Closing after merge.

mrbean-bremen:
Forgot to fix ResolvePath() - reopen.

mrbean-bremen:
Back to closed - no more known issues.

PR #84, opened 2016-04-11
Fixed support for drive letter

mrbean-bremen:
- fixed behavior of NormalizePath() and JoinPaths() with drive letters
- shall fix issue #83

### DIFF
--- a/pyfakefs/fake_filesystem.py
+++ b/pyfakefs/fake_filesystem.py
@@ -371,6 +371,7 @@ class FakeFilesystem(object):
     if path_separator != os.sep:
       self.alternative_path_separator = None
     self.is_case_sensitive = not _is_windows and sys.platform != 'darwin'
+    self.supports_drive_letter = _is_windows
     self.root = FakeDirectory(self.path_separator)
     self.cwd = self.root.name
     # We can't query the current value without changing it:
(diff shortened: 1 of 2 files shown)


pyfakefs/fake_filesystem.py
```diff
@@ -371,6 +371,7 @@ def __init__(self, path_separator=os.path.sep):
     if path_separator != os.sep:
       self.alternative_path_separator = None
     self.is_case_sensitive = not _is_windows and sys.platform != 'darwin'
+    self.supports_drive_letter = _is_windows
```
jmcgeheeiv: Should this be private attribute `self._supports_drive_letter`?

pyfakefs/fake_filesystem.py
```diff
@@ -371,6 +371,7 @@ def __init__(self, path_separator=os.path.sep):
     if path_separator != os.sep:
       self.alternative_path_separator = None
     self.is_case_sensitive = not _is_windows and sys.platform != 'darwin'
```
jmcgeheeiv: Should this be a private attribute?

pyfakefs/fake_filesystem.py
```diff
@@ -371,6 +371,7 @@ def __init__(self, path_separator=os.path.sep):
     if path_separator != os.sep:
       self.alternative_path_separator = None
```
jmcgeheeiv: Should this be a private attribute?

mrbean-bremen:
I use is_case_sensitive and alternative_path_separator in tests to be able to test this under Linux (to minimize the Windows-only tests as there is no Windows test system). I can make them private and access them from the tests anyway, don't know what your policy is here, but I would prefer the tests not to be dependent on Windows.

Issue #86, opened 2016-04-14
Support for file system disk size


mrbean-bremen:
I would like to have the possibility to set the available disk space in the file system to be able to test code under out of space conditions or code depending on the remaining disk space.
I'm thinking about putting together something very basic (at least for the start) in this order:
- set the file system size regardless of paths 
- calculate remaining disk space solely based in the file size (there is already that handy feature of large files, that allows you to virtually fill the disk)
- throw IO exception on trying to create a file that is too large for the remaining disk space
- allow to access free disk space (this allows at least for test-aware production code for getting the remaining disk size)
- implement shutil.disk_usage (better method, available unfortunately only since Python 3.3)
- add basic support for os.statvfs (all versions, but Unix only)

Not so sure about:
- replacement for the Windows method of getting disk space (which is something like ctypes.windll.kernel32.GetDiskFreeSpaceExW)
- adding support for path-specific sizes

What do you think?

mrbean-bremen:
For the record: implemented in PR #87 except the support for the os specific functions (os.statvfs and GetDiskFreeSpaceExW) and the path-specific part (which I won't implement until there is real need).
I'm not sure if and how to implement the os-specific functions and if this is needed (for me, the current state would be sufficient). Any thoughts?

mrbean-bremen:
Works ok for me as is now, closing this. Reopen at will if you need more...

PR #87, opened 2016-04-16
Adding support for filesystem size

mrbean-bremen:
- added parameter to set the total size of the filesystem
- keep track of used size for added and deleted files
- raise IOError for insufficient space on creating or resizing a file
- added shutil.disk_usage() implementation
- see issue #86

### DIFF
--- a/fake_filesystem_test.py
+++ b/fake_filesystem_test.py
@@ -3293,15 +3293,19 @@ class DiskSpaceTest(TestCase):
     self.assertEqual((100, 5, 95), self.filesystem.GetDiskUsage())
 
   def testFileSystemSizeAfterAsciiStringFileCreation(self):
-    self.filesystem.CreateFile('/foo/bar', contents=u'complicated')
+    if sys.version_info < (3, 0):
+      contents = u'complicated'
+    else:
+      contents = 'complicated'
+    self.filesystem.CreateFile('/foo/bar', contents=contents)
     self.assertEqual((100, 11, 89), self.filesystem.GetDiskUsage())
 
   def testFileSystemSizeAfter2ByteUnicodeStringFileCreation(self):
-    self.filesystem.CreateFile('/foo/bar', contents=u'сложно')
+    self.filesystem.CreateFile('/foo/bar', contents='сложно')
     self.assertEqual((100, 12, 88), self.filesystem.GetDiskUsage())
 
   def testFileSystemSizeAfter3ByteUnicodeStringFileCreation(self):
-    self.filesystem.CreateFile('/foo/bar', contents=u'複雑')
+    self.filesystem.CreateFile('/foo/bar', contents='複雑')
     self.assertEqual((100, 6, 94), self.filesystem.GetDiskUsage())
 
   def testFileSystemSizeAfterFileDeletion(self):
--- a/fake_filesystem_test.py
+++ b/fake_filesystem_test.py
@@ -3189,5 +3189,65 @@ class AlternativePathSeparatorTest(TestCase):
     self.assertTrue(self.filesystem.Exists('?foo?bar?xyzzy?plugh'))
 
 
+class DiskSpaceTest(TestCase):
+  def setUp(self):
+    self.filesystem = fake_filesystem.FakeFilesystem(path_separator='/', total_size=1024)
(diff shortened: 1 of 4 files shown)


fake_filesystem_test.py
```diff
@@ -3250,5 +3250,65 @@ def testExistsWithMixedSeparators(self):
     self.assertTrue(self.filesystem.Exists('?foo?bar?xyzzy?plugh'))
 
 
+class DiskSpaceTest(TestCase):
+  def setUp(self):
+    self.filesystem = fake_filesystem.FakeFilesystem(path_separator='/', total_size=1024)
+
+  def testFileSystemSizeAfterFileCreation(self):
```
jmcgeheeiv: File contents are commonly strings.

Please add a test trying the ASCII string 'xyzzy', I suppose that this should consume:

> 5 characters + null terminator = 6 bytes

Then add a test trying the Unicode string '複雑' which means "complicated" in Japanese and Chinese.  I would think this would consume:

> 3 bytes for 複 + 3 bytes for 雑 + null terminator = 7 bytes

pyfakefs/fake_filesystem.py
```diff
@@ -248,38 +255,76 @@ def SetLargeFileSize(self, st_size):
...
       contents = Hexlified(contents)
 
-    self.st_ctime = time.time()
-    self.st_mtime = self._st_ctime
-    self.st_size = len(contents)
+    if self.contents:
+      self.SetSize(0)
+    current_size = self.st_size or 0
     self.contents = contents
+    self.st_size = len(self.contents)
```
jmcgeheeiv: This gives the wrong answer for non-ascii strings:

    >>> len('複雑')
    2

Maybe you need something like this:

    >>> len('複雑'.encode("utf8"))
    6

I tried actually writing this to a file, and Linux `ls -l` says the file size is 6.

mrbean-bremen:
Are you sure about the null terminator? AFAIK, this is not added automatically if writing python strings. About the unicode strings: you are right, I will add a test, and fix the code, currently it is just taking len(contents), which will not work correctly with unicode strings. 
But I had thought that unicode strings in both Python 2 and 3 are written as UCS-2 (or UCS-4, but your nice Kanji shall fit into the 16 bit code point range), which would mean 2 bytes per character. Anyway, I will check this...

mrbean-bremen:
Ok, didn't see your second post as I answered the first one. The problem I see here is that you encoded the string in UTF-8 before writing it to the disk, which is not a given, as I understand it. But I think I have to better understand the code first - especially how reading and writing is done in the fake file system. I will ask you if I need some help there... thanks for your support.

mrbean-bremen:
Ok, I made a few tests - you are right about UTF-8, I didn't know that. Under Python 2, strings are indeed written as UTF-8. Under Python 3, if the file is opened binary, you have to provide a bytestring (e.g. do the encoding yourself, as you did), otherwise it depends on your text file encoding. I will have a go at this somewhere in the next days.

jmcgeheeiv:
I read/write Japanese and I am relatively familiar with non-ASCII, so if you need anything I am at your service.

Indeed I was wrong about the null terminator.  I removed my comments suggesting it might be necessary.

mrbean-bremen:
Ok, I hopefully fixed this - please check if I shall add more tests. And if I ever want to learn Japanese, I will get back to you :)

fake_filesystem_test.py
```diff
@@ -3276,5 +3276,84 @@ def testSplitPath(self):
...
+
+  def testFileSystemSizeAfterBinaryFileCreation(self):
+    self.filesystem.CreateFile('/foo/bar', contents=b'xyzzy')
+    self.assertEqual((100, 5, 95), self.filesystem.GetDiskUsage())
+
+  def testFileSystemSizeAfterAsciiStringFileCreation(self):
+    self.filesystem.CreateFile('/foo/bar', contents='complicated')
+    self.assertEqual((100, 11, 89), self.filesystem.GetDiskUsage())
+
+  def testFileSystemSizeAfter2ByteUnicodeStringFileCreation(self):
```
jmcgeheeiv: Two byte UTF-8, very cool.

PR #88, opened 2016-04-19
Fixed behavior of ResolvePath() for Windows paths

mrbean-bremen:
see issue #83


jmcgeheeiv:
With all your fixes, it is getting to be time for a release.  An "enhanced Windows support" release.

Please let me know when you think it is time.  When it's really time; no rush.

mrbean-bremen:
This is a good idea - I would like it. I would still wait some time, as there may pop up more issues, maybe in a couple of weeks or so...

PR #89, opened 2016-04-19
Fixed GetPathComponents() with Windows path

mrbean-bremen:
- had used _IsRootPath() once too often

### DIFF
--- a/fake_filesystem_test.py
+++ b/fake_filesystem_test.py
@@ -1965,6 +1965,13 @@ class FakePathModuleTest(TestCase):
     self.assertEqual(self.path.join(basedir, file_components[1]),
                      self.path.abspath(file_components[1]))
 
+  def testAbsPathWithDriveComponent(self):
+    self.filesystem.supports_drive_letter = True
+    self.filesystem.cwd = 'C:/foo'
+    self.assertEqual('C:/foo/bar', self.path.abspath('bar'))
+    self.assertEqual('C:/foo/bar', self.path.abspath('C:bar'))
+    self.assertEqual('C:/foo/bar', self.path.abspath('/foo/bar'))
+
   def testRelpath(self):
     path_foo = '/path/to/foo'
     path_bar = '/path/to/bar'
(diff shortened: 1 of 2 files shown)


fake_filesystem_test.py
```diff
@@ -3279,6 +3279,10 @@ def testCharactersBeforeRootIgnoredInJoinPaths(self):
   def testResolvePath(self):
     self.assertEqual('c:/foo/bar', self.filesystem.ResolvePath('c:/foo/bar'))
 
+  def testGetPathComponents(self):
+    self.assertEqual(['c:', 'foo', 'bar'], self.filesystem.GetPathComponents('c:/foo/bar'))
+    self.assertEqual(['c:'], self.filesystem.GetPathComponents('c:'))
+
```
jmcgeheeiv: Is there any kind of degenerate but still legal path that should be tried, such as `c:foo/bar` or `c:\foo/bar`?  I do not know if `c:foo/bar` is in fact legal.

mrbean-bremen:
I will check this tomorrow - too late for today. C:foo/bar is legal, but practically not used - may be this case will indeed fail. It is a path relative to the cwd on C: (as on Windows there is a separate current directory on each drive).

jmcgeheeiv:
Ah.  You taught me something.  Once I was using Windows command line, and the directory name acted quite peculiarly.  Maybe it was the difference between `C:` and `C:/`.  I did not think they were different--especially not *that* different.  I'll have to watch out for this next time.

mrbean-bremen:
Ok, I added a couple of tests, and the code for join() turned out a little complicated (I knew why I didn't want to add full drive support...) Anyway, in the end I threw it away and mostly copied the Windows-specific code for os.path.join() from the Python implementation, which by definition shall do the Right Thing. I'm not sure if this is a good idea - I will check it in now and you can decide. 
I also didn't add support for UNC paths (hopefully nobody needs them), so the Windows implementation is far from complete.

mrbean-bremen:
This did not work obviously - I better make a new PR and close this one.

PR #90, opened 2016-04-20
Some drive-letter related fixes

mrbean-bremen:
- use adapted Windows specific Python join() with drive letter support
- fixed abspath() wih regard to drive letters
- fixed previously introduced bug in GetPathComponents()
- implemented splitdrive() to be able to test it under Linux


mrbean-bremen:
Ok, still fails under Linux - closing. Time to sleep...

PR #92, opened 2016-04-21
Drive letter fixes

mrbean-bremen:
- several fixes for drive letter support
- initial (very basic) support for UNC paths


mrbean-bremen:
I think I'm done for the moment with the drive support; I added some basic UNC path handling (only since  2.7.8, as before that UNC paths were only partly supported in Python, and I don't think we need to add this partial support to 2.6 only).

PR #94, opened 2016-04-26
Drive letter handling fix

mrbean-bremen:
- correctly handle drive root in SplitPath() (hopefully last fix )


mrbean-bremen:
I did some more tests, found the remaining bug I fixed in this PR, but otherwise it seems to work for me so far, though I'm sure there are more issues lurking somewhere... Anyway, I think I'm done for now. Maybe after CausticYarn has checked if this works for him, too, you may think about a new version.
Thanks for your patience...

mrbean-bremen:
Accidentally closed - pressed the wrong button... reopen.

Issue #75, opened 2016-03-25
os.link is implemented to make symlinks instead of hardlinks


sruggier:
I was trying to use the atomicwrites package under pyfakefs, and after adding fsync support (see #73), the next issue I noticed was that the fake version of os.link is aliased to the fake os.symlink, which has very different behaviour in cases where the original file is renamed or deleted:
```
  # TODO: Link doesn't behave like os.link, this needs to be fixed properly.
  link = symlink
```

jmcgeheeiv:
Excellent observation.  It would be fabulous if you could issue a pull request for this, too.

sruggier:
Yeah, I've been working on it on and off. I have a working implementation that I'll make a PR out sometime in the next few days.

jmcgeheeiv:
Considering this fixed by #95 and #97.

PR #95, opened 2016-04-28
Implement hard link support

sruggier:
This change adds support for hard links, and tests to go along with
that.

There is a bit of ugliness in this change around the use of the name
field on FakeFile instances. Support for hard links means that the same
file can have different names under different directories, so it
fundamentally doesn't make sense for FakeFile instances to have a name
associated with them.

However, I didn't want to have to perform a mass refactoring as a
prerequisite to making this change, so it instead just sets the filename
as needed to work with the existing implementation of AddObject. This
has the unavoidable side effect of changing the name associated with the
original file as well, since they share the same FakeFile instance.

It looks like this is probably safe to do now, but to properly ensure
that there aren't any problems related to this, the whole codebase
should be refactored to remove the name field from FakeFile, and replace
it with a suitable alternative. It may require a wrapper class that
helps callers keep track of a file's name while preventing misuse of
that information by direct users of FakeFile.

### DIFF
--- a/pyfakefs/fake_filesystem.py
+++ b/pyfakefs/fake_filesystem.py
@@ -210,10 +210,10 @@ class FakeFile(object):
       self.st_size = len(contents)
     else:
       self.st_size = 0
+    self.st_nlink = 1
     # Non faked features, write setter methods for fakeing them
     self.st_ino = None
     self.st_dev = None
-    self.st_nlink = None
     self.st_uid = None
     self.st_gid = None
     # shall be set on creating the file from the file system to get access to fs available space
@@ -436,6 +436,11 @@ class FakeDirectory(FakeFile):
     """
     if pathname_name in self.contents and self.filesystem:
       self.filesystem.ChangeDiskUsage(-self.contents[pathname_name].GetSize())
+
+    entry = self.contents[pathname_name]
+    entry.st_nlink -= 1
+    assert entry.st_nlink >= 0
+
     del self.contents[pathname_name]
 
   def GetSize(self):
@@ -1244,6 +1249,49 @@ class FakeFilesystem(object):
     resolved_file_path = self.ResolvePath(file_path)
     return self.CreateFile(resolved_file_path, st_mode=stat.S_IFLNK | PERM_DEF,
                            contents=link_target)
+
+  def CreateHardLink(self, old_path, new_path):
+    """Create a hard link at new_path, pointing at old_path.
+
+    Args:
+      old_path: an existing link to the target file
+      new_path: the destination path to create a new link at
+
+    Returns:
+      the FakeFile object referred to by old_path
+
+    Raises:
+      OSError:  if something already exists at new_path
+      OSError:  if the parent directory doesn't exist
+    """
+    new_path_normalized = self.NormalizePath(new_path)
+    if self.Exists(new_path_normalized):
+      raise IOError(errno.EEXIST,
+                    'File already exists in fake filesystem',
+                    new_path)
+
+    new_parent_directory, new_basename = self.SplitPath(new_path_normalized)
+    if not new_parent_directory:
+      new_parent_directory = self.cwd
+
+    if not self.Exists(new_parent_directory):
+      raise OSError(errno.ENOENT, 'No such fake directory',
+                    new_parent_directory)
+
+    # Retrieve the target file
+    try:
+      old_file = self.GetObject(old_path)
+    except:
+      raise OSError(errno.ENOENT,
+                    'No such file or directory in fake filesystem',
+                    old_path)
+
+    old_file.st_nlink += 1
+
+    # abuse the name field to control the filename of the newly created link
+    old_file.name = new_basename
+    self.AddObject(new_parent_directory, old_file)
+    return old_file
 
   def __str__(self):
     return str(self.root)
@@ -2169,9 +2217,8 @@ class FakeOsModule(object):
     """
     self.filesystem.CreateLink(path, link_target)
 
-  # pylint: disable-msg=C6002
-  # TODO: Link doesn't behave like os.link, this needs to be fixed properly.
-  link = symlink
+  def link(self, oldpath, newpath):
+      self.filesystem.CreateHardLink(oldpath, newpath)
 
   def fsync(self, file_des):
     """Perform fsync for a fake file (in other words, do nothing).
--- a/fake_filesystem_test.py
+++ b/fake_filesystem_test.py
@@ -1740,6 +1740,88 @@ class FakeOsModuleTest(TestCase):
     self.assertTrue(self.os.path.lexists(file_path))
     self.assertTrue(self.os.path.exists(file_path))
 
+  # hard link related tests
+
+  def testLinkBogus(self):
+    # trying to create a link from a non-existent file should fail
+    self.assertRaises(OSError,
+                     self.os.link, '/nonexistent_source', '/link_dest')
+
+  def testLinkDelete(self):
+    fake_open = fake_filesystem.FakeFileOpen(self.filesystem)
+
+    file1_path = 'test_file1'
+    file2_path = 'test_file2'
+    contents1 = 'abcdef'
+    # Create file
+    self.filesystem.CreateFile(file1_path, contents=contents1)
+    # link to second file
+    self.os.link(file1_path, file2_path)
+    # delete first file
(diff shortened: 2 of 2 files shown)


sruggier:
I forgot to mention: this fixes #75.

pyfakefs/fake_filesystem.py
```diff
@@ -1245,6 +1250,49 @@ def CreateLink(self, file_path, link_target):
...
+    Returns:
+      the FakeFile object referred to by old_path
+
+    Raises:
+      OSError:  if something already exists at new_path
+      OSError:  if the parent directory doesn't exist
+    """
+    new_path_normalized = self.NormalizePath(new_path)
+    if self.Exists(new_path_normalized):
+      raise IOError(errno.EEXIST,
```
jmcgeheeiv: It's not necessarily a file.  Perhaps,

```"Fake file system object already exists at path", new_path ```

Comment (sruggier) [inline pyfakefs/fake_filesystem.py]: Yeah, fair enough. I checked what `ln` reports, for inspiration, and they make the same mistake. I think "fake filesystem object" might be confusing for some peope. How about `Fake file or directory already exists at path`?

**Comment by jmcgeheeiv** [comment]
Let's go ahead and follow the lead of `ln`.  You are right.  "File or directory' is better than "Fake file system object".

I'm still trying to wrap my head around that `old_file.name = new_name` thing.

**Comment by sruggier** [comment]
From my attempt to review the codebase, I think the worst possible ramification right now is that some error messages could use the wrong filename, but that could change if someone tries to use that field for something more concrete than diagnostic output.

File: pyfakefs/fake_filesystem.py [hunk 2]
```diff
@@ -1245,6 +1250,49 @@ def CreateLink(self, file_path, link_target):
...
+    except:
+      raise OSError(errno.ENOENT,
+                    'No such file or directory in fake filesystem',
+                    old_path)
+
+    old_file.st_nlink += 1
+
+    # abuse the name field to control the filename of the newly created link
+    old_file.name = new_basename
+    self.AddObject(new_parent_directory, old_file)
```
mrbean-bremen: This will currently also increase the file system size by the linked file size, as st_nlink is not taken into account in AddEntry(). I may fix this after the PR is merged back, as I added the file system size stuff only recently and could do some minor code cleanup there anyway.

sruggier: Right, the FS size stuff wasn't around when I first implemented this. I noticed it when I rebased, there was a merge conflict in the unlink implementation, but I didn't do anything beyond just resolving that conflict. It's probably not critical to fix this issue prior to merging, though, given that both the file sizing and hard link support are new.

You can also play with these commits before they're merged, if you add a line like this into your git config:
```
[remote "origin"]
...
	fetch = +refs/pull/*/head:refs/remotes/origin/pr/*
```

That will cause git fetch to make all open PRs available under origin/pr/<number>.

It would make sense for me to update this myself, but I've been pretty busy in the last few weeks, and I'm not sure when I'd find time to get it done.

mrbean-bremen: You are right that it's not critical, I can play with this this over the weekend if I find the time. I do not expect large changes to get it working, I will let you know when I'm done.

mrbean-bremen: Ok, the following patch to your PR fixes the problem (and contains some minor refactoring regarding ChangeDiskUsage()):
```
 pyfakefs/fake_filesystem.py | 48 ++++++++++++++++++++++-----------------------
 1 file changed, 23 insertions(+), 25 deletions(-)

diff --git a/pyfakefs/fake_filesystem.py b/pyfakefs/fake_filesystem.py
index 8e6b731..b0e9f59 100644
--- a/pyfakefs/fake_filesystem.py
+++ b/pyfakefs/fake_filesystem.py
@@ -267,12 +267,8 @@ class FakeFile(object):
                     self.name)
     if self.st_size:
       self.SetSize(0)
-    if self.filesystem and self.filesystem.total_size is not None:
-      if self.filesystem.GetDiskUsage().free < st_size:
-        raise IOError(errno.ENOSPC,
-                      'Fake file system: disk is full, failed to set file size to %r bytes' % st_size,
-                      self.name)
-      self.filesystem.ChangeDiskUsage(st_size)
+    if self.filesystem:
+      self.filesystem.ChangeDiskUsage(st_size, self.name)
     self.st_size = st_size
     self.contents = None
 
@@ -305,12 +301,8 @@ class FakeFile(object):
     current_size = self.st_size or 0
     self.contents = contents
     self.st_size = st_size
-    if self.filesystem and self.filesystem.total_size is not None:
-      if self.filesystem.GetDiskUsage().free < self.st_size - current_size:
-        raise IOError(errno.ENOSPC,
-                      'Fake file system: disk is full, failed to set contents to %r bytes' % self.st_size,
-                      self.name)
-      self.filesystem.ChangeDiskUsage(self.st_size - current_size)
+    if self.filesystem:
+      self.filesystem.ChangeDiskUsage(self.st_size - current_size, self.name)
     self.epoch += 1
 
   def SetContents(self, contents):
@@ -339,6 +331,7 @@ class FakeFile(object):
 
     Raises:
       IOError: if the st_size arg is not a non-negative integer
+               or if st_size exceeds the available file system space
     """
 
     if not isinstance(st_size, int) or st_size < 0:
@@ -348,12 +341,8 @@ class FakeFile(object):
                     self.name)
 
     current_size = self.st_size or 0
-    if self.filesystem and self.filesystem.total_size is not None:
-      if self.filesystem.GetDiskUsage().free < st_size - current_size:
-        raise IOError(errno.ENOSPC,
-                      'Fake file system: disk is full, failed to set file size to %r bytes' % st_size,
-                      self.name)
-      self.filesystem.ChangeDiskUsage(st_size - current_size)
+    if self.filesystem:
+      self.filesystem.ChangeDiskUsage(st_size - current_size, self.name)
     if self.contents:
       if st_size < current_size:
         self.contents = self.contents[:st_size]
@@ -410,8 +399,8 @@ class FakeDirectory(FakeFile):
       path_object:  FakeFile instance to add as a child of this directory
     """
     self.contents[path_object.name] = path_object
-    if self.filesystem:
-      self.filesystem.ChangeDiskUsage(path_object.GetSize())
+    if self.filesystem and path_object.st_nlink == 1:
+      self.filesystem.ChangeDiskUsage(path_object.GetSize(), path_object.name)
 
   def GetEntry(self, pathname_name):
     """Retrieves the specified child file or directory.
@@ -434,10 +423,10 @@ class FakeDirectory(FakeFile):
     Raises:
       KeyError: if no child exists by the specified name
     """
-    if pathname_name in self.contents and self.filesystem:
-      self.filesystem.ChangeDiskUsage(-self.contents[pathname_name].GetSize())
-
     entry = self.contents[pathname_name]
+    if self.filesystem and entry.st_nlink == 1:
+      self.filesystem.ChangeDiskUsage(-entry.GetSize(), pathname_name)
+
     entry.st_nlink -= 1
     assert entry.st_nlink >= 0
 
@@ -498,14 +487,23 @@ class FakeFilesystem(object):
       return DiskUsage(self.total_size, self.used_size, self.total_size - self.used_size)
     return DiskUsage(1024*1024*1024*1024, 0, 1024*1024*1024*1024)
 
-  def ChangeDiskUsage(self, usage_change):
+  def ChangeDiskUsage(self, usage_change, file_path):
     """Changes the used disk space by the given amount.
 
     Args:
       usage_change: number of bytes added to the used space
                     if negative, the used space will be decreased
+
+      file_path: the path of the object needing the disk space
+
+    Raises:
+      IOError: if usage_change exceeds the free file system space
     """
-    if self.used_size is not None:
+    if self.total_size is not None:
+      if self.total_size - self.used_size < usage_change:
+        raise IOError(errno.ENOSPC,
+                      'Fake file system: disk is full, failed to add %r bytes' % usage_change,
+                      file_path)
       self.used_size += usage_change
 
   def SetIno(self, path, st_ino):
```

mrbean-bremen: I had a second look at os.link and noticed that since Python 3.2 it also supports hard links under Windows - I wasn't aware of this, as I'm still mostly using Python 2.7. I may have a look at this later, but I think it is easier to first merge back your changes.

jmcgeheeiv:
@mrbean-bremen, could I ask you to issue a PR with the changes in your patch?

PR #96, opened 2016-04-30
Fixed handling of file system size with hard links

mrbean-bremen:
- added check for st_nlink in file system size change
- moved raising out of space IOError to single location
- added test, disabled some tests under Windows

### DIFF
--- a/pyfakefs/fake_filesystem.py
+++ b/pyfakefs/fake_filesystem.py
@@ -267,12 +267,8 @@ class FakeFile(object):
                     self.name)
     if self.st_size:
       self.SetSize(0)
-    if self.filesystem and self.filesystem.total_size is not None:
-      if self.filesystem.GetDiskUsage().free < st_size:
-        raise IOError(errno.ENOSPC,
-                      'Fake file system: disk is full, failed to set file size to %r bytes' % st_size,
-                      self.name)
-      self.filesystem.ChangeDiskUsage(st_size)
+    if self.filesystem:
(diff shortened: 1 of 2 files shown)


pyfakefs/fake_filesystem.py
```diff
@@ -498,14 +487,23 @@ def GetDiskUsage(self):
       return DiskUsage(self.total_size, self.used_size, self.total_size - self.used_size)
     return DiskUsage(1024*1024*1024*1024, 0, 1024*1024*1024*1024)
 
-  def ChangeDiskUsage(self, usage_change):
+  def ChangeDiskUsage(self, usage_change, file_path):
```
jmcgeheeiv: Nice refactoring.

jmcgeheeiv:
Earlier you mentioned that in some version, Python started to support hard links on Windows.  So that this useful fact does not get forgotten, could I ask you to enter an issue requesting this?

mrbean-bremen:
I just had a look at the problem with the failing tests, and it turned out that NormalizeCase() didn't work with hard links correctly because it relied on the saved file name (which in the case of a hard link is the original name). I wanted to append the fix to the PR and only just noticed that you already merged it, so I made a new one.
As the fix by sruggier also works on Windows now, we don't need an issue. The only difference now is that hard links in fakefs are also supported in Python 2.7, where they are not supported in the real code - I'm not sure if this is relevant, in that case we have to add a version check in the hard link code.

PR #97, opened 2016-04-30
Fixed NormalizeCase() with hard links

mrbean-bremen:
- changed to not rely on file name member as this may be incorrect in
case of hard links
- re-enabled tests under Windows


mrbean-bremen:
For the record (as the original discussion is lost in some other PR): this is related to issue #75 and fixes the behavior in Windows.
Remaining is the issue that symlinks and hardlinks in Windows are only supported since Python 3.2, while fakefs supports them in all versions - I'm not sure if this a real problem, please check if this shall be changed.

Issue #99, opened 2016-05-05
Renaming directory does not adapt st_nlink


mrbean-bremen:
The following test:
```
    filesystem = fake_filesystem.FakeFilesystem(path_separator='/')
    fake_os = fake_filesystem.FakeOsModule(filesystem)
    filesystem.CreateDirectory('/foo')
    fake_os.rename('/foo', '/bar')
    fake_os.rmdir('/bar')
```
results in:
```
  File "pyfakefs\pyfakefs\fake_filesystem.py", line 1975, in rmdir
    self.filesystem.RemoveObject(target_directory)
  File "pyfakefs\pyfakefs\fake_filesystem.py", line 1134, in RemoveObject
    target_directory.RemoveEntry(basename)
  File "pyfakefs\pyfakefs\fake_filesystem.py", line 431, in RemoveEntry
    assert entry.st_nlink >= 0
AssertionError
```

mrbean-bremen:
Fixed with #101

PR #100, opened 2016-05-06
Simplified rename

mrbean-bremen:
- change rename to change the existing object instead of creating a new
one
- initialize st_link with 0 and increase it in AddEntry() (mirrors
behavior in RemoveEntry())
- fixes issue #99


mrbean-bremen:
Tests fail - closing.

PR #101, opened 2016-05-06
Rename fix

mrbean-bremen:
- change rename to change the existing object instead of creating a new
one
- initialize st_link with 0 and increase it in AddEntry() (mirrors
behavior in RemoveEntry())
- fixes issue #99


mrbean-bremen:
Please check if it is ok to simplify renaming this way, or if I overlooked something why the old implementation was needed.
Also, if you want to merge this, better merge the 2 commits - I did this locally, but it got messed up with the existing pull request.

PR #107, opened 2016-06-08
Added support for mount points

### LINKED ISSUE #25: Improve handling of st_ino
```
Currently pyfakefs handles st_ino as something optional, which is probably OK 
for many scenarios.

But it would be really nice, if st_ino would be handled properly. 

I'm working on a sort of file syncing application, which needs to detect all FS 
modifications. Comparing st_ino of two files/dirs (e.g. a file from an old FS 
snapshot and a file from a new FS snapshot) is a typical and reliable way to 
detect that both files/dirs correspond to the same object on the file system. 
If st_ino values are the same, then it is the same object (at least on most 
UNIX file systems). 

Therefore, I'd like to propose the following:
1) When a new file or directory is created, it should get a new unique st_ino 
value

2) When objects (files or dirs) are moved inside the same file system, their 
st_ino fields should be preserved.
    (BTW, fake_filesystem_shutil currently tries to preserve most st_* stats for 
files, but st_ino is not included into this set of preserved fields. And for
directory moves, no st_* stats are preserved at all for some reason)

3) Optional: currently fake_filesystem_shutil.move uses copying to move files 
and directories. This does not scale for big nested directories, because it 
needs to move each nested sub-component separately. May be it could use 
"rename" whenever possible? It could provide significant speed-ups.

```

Original issue reported on code.google.com by `romixlev` on 6 Jan 2014 at 12:15

mrbean-bremen:
- added AddMountPoint() together with support for st_ino and st_dev
- adapted GetDiskUsage() and ChangeDiskUsage() to use the file system given by path, added SetDiskUsage() convenience function
- added FakeOsModule.ismount() implementation
- moved rename implementation to FakeFilesystem.RenameObject(), adapted
it and shutil.move() to correctly behave with same and different file systems
- added auto-mount for Windows drives and UNC paths and some related fixes 
- adapted/added tests
Fixes #25

### DIFF
--- a/fake_filesystem_test.py
+++ b/fake_filesystem_test.py
@@ -113,8 +113,8 @@ class FakeDirectoryUnitTest(TestCase):
     filesystem = fake_filesystem.FakeFilesystem(path_separator='/')
     fake_os = fake_filesystem.FakeOsModule(filesystem)
     file_path = 'some_file1'
-    filesystem.CreateFile(file_path, contents='contents here1', inode=42)
-    self.assertEqual(42, fake_os.stat(file_path)[stat.ST_INO])
+    filesystem.CreateFile(file_path, contents='contents here1')
+    self.assertLess(0, fake_os.stat(file_path)[stat.ST_INO])
 
     file_obj = filesystem.GetObject(file_path)
     file_obj.SetIno(43)
@@ -124,8 +124,8 @@ class FakeDirectoryUnitTest(TestCase):
     filesystem = fake_filesystem.FakeFilesystem(path_separator='/')
     fake_os = fake_filesystem.FakeOsModule(filesystem)
     dirpath = 'testdir'
-    filesystem.CreateDirectory(dirpath, inode=42)
-    self.assertEqual(42, fake_os.stat(dirpath)[stat.ST_INO])
+    filesystem.CreateDirectory(dirpath)
+    self.assertLess(0, fake_os.stat(dirpath)[stat.ST_INO])
 
     dir_obj = filesystem.GetObject(dirpath)
     dir_obj.SetIno(43)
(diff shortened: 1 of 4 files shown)


fake_filesystem_test.py
```diff
@@ -2209,6 +2230,42 @@ def testIslink(self):
...
+    self.assertTrue(self.path.ismount('z:/'))
+    self.filesystem.AddMountPoint('/mount')
+    self.assertTrue(self.path.ismount('/mount'))
+    self.assertTrue(self.path.ismount('/mount/'))
+
+  @unittest.skipIf(sys.version_info < (2, 7, 8), 'UNC path support since Python 2.7.8')
+  def testIsmountWithUncPaths(self):
+    self.filesystem.supports_drive_letter = True
+    self.assertTrue(self.path.ismount('//a/'))
+    self.assertTrue(self.path.ismount('//a/b'))
```
jmcgeheeiv: How did //a/b become a mount point?

pyfakefs/fake_filesystem.py
```diff
@@ -180,6 +180,8 @@ class FakeFile(object):
        st_atime: the time.time() timestamp when the file was last accessed.
        st_mtime: the time.time() timestamp when the file was last modified.
        st_size: the size of the file
+       st_ino: the inode number - a unique number identifying the file
+       st_dev: a unique number identifying the (fake) file system device the file belongs to
 
      Other attributes needed by os.stat are assigned default value of None
       these include: st_ino, st_dev, st_nlink, st_uid, st_gid
```
jmcgeheeiv: Remove "st_ino, st_dev" from this list of unimplemented attributes.

fake_filesystem_test.py
mrbean-bremen: Via SplitDrive(). This also splits off UNC paths (since 2.7.8).

pyfakefs/fake_filesystem.py
mrbean-bremen: Done.

jmcgeheeiv:
Forgive me, I forgot to mention that this is TOTALLY AWESOME!  Mounting disks...who would have ever thought?

Issue #103, opened 2016-05-10
zipfile.ZipFile does not create file in fake file system on Windows with Python 2.7.9


borismod:
I am using zipfile package for creating zip file. The below test fails. Is there any way to make it work?

```Python
import os
import zipfile
from pyfakefs import fake_filesystem_unittest


class TestPackageBuilder(fake_filesystem_unittest.TestCase):
    def setUp(self):
        self.setUpPyfakefs()
   
    def test_zip(self):
        self.fs.CreateFile(u'C:\\work\\readme.txt', contents='')
        zip_file_path = os.path.join(u'c:\\work', 'some_file.zip')
        with zipfile.ZipFile(zip_file_path, 'w') as zip_file:
            zip_file.write(u'C:\\work\\readme.txt')

        self.assertFileExists(u'c:\\work\\some_file.zip')

    def assertFileExists(self, file_path):
        self.assertTrue(os.path.exists(file_path), msg='File {0} does not exist'.format(file_path))

```

mrbean-bremen:
This may be an issue with case-sensitivity, which is fixed in the master branch, but not yet in the PyPi package. Try to use the same case everywhere (there is both 'C:' and 'c:' used), or try to install the master branch:
pip install -U https://github.com/jmcgeheeiv/pyfakefs/archive/master.zip
and check if it works.

borismod:
Still fails with master and with same case:

```Python
    def test_zip(self):
        self.fs.CreateFile(u'c:\\work\\readme.txt', contents='')
        zip_file_path = os.path.join(u'c:\\work', 'some_file.zip')
        with zipfile.ZipFile(zip_file_path, 'w') as zip_file:
            zip_file.write(u'c:\\work\\readme.txt')

        self.assertFileExists(u'c:\\work\\some_file.zip')
```

mrbean-bremen:
That's strange... I just tried it locally (with the master branch) and it works fine for me (both versions).

borismod:
I am running on Win10, Pyton 2.7 if it matters

mrbean-bremen:
I am currently run under Python 2.7 and Windows 7 (at work), which should not matter, but I may test it at Windows 10 at home in the evening. Do you use a specific test runner or environment?

borismod:
Nope, I am running on PyCharm Pro 5.0.4, which should not matter

mrbean-bremen:
[...]

PR #109, opened 2016-06-12
Fake io open

### LINKED ISSUE #70: python3 - io.open not redirected to fake file system
```python 
def setUp(self):
    self.setUpPyfakefs()
    # Fix for python 3+; except 3.2
    #io.open = open

def test_willRedirectIoOpen(self):
     os.makedirs('install')
     file = os.path.join('install', 'sample.txt')
        with io.open(file, 'w'):
            pass
    self.assertTrue(os.path.exists(file))
```
this causes the following error: IOError: [Errno 2] No such file or directory: 'install\\sample.txt'

Monkey patching io.open to open after setting up the fake file system seems to fix the problem for all python versions with the exception of 3.2

mrbean-bremen:
- redirects to FakeFileOpen
- fixes #70 and #103

### DIFF
--- a/fake_filesystem_unittest_test.py
+++ b/fake_filesystem_unittest_test.py
@@ -18,7 +18,7 @@
 """
 Test the :py:class`pyfakefs.fake_filesystem_unittest.TestCase` base class.
 """
-
+import io
 import os
 import glob
 import shutil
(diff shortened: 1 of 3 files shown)


jmcgeheeiv:
Looks good.  I'll merge this after you fix the minor issue in #107.

jmcgeheeiv:
Another fine contribution from @mrbean-bremen.

borismod:
@jmcgeheeiv thanks for fixing it. Was it released to pypi? What version?

PR #137, opened 2016-11-19
Added auto doc generation

mrbean-bremen:
- added sphinx configuration to create documentation from docstrings
- fixed some sphinx warnings in docstrings
- added script that recreates documentation after each commit and commit
it to gh-pages


mrbean-bremen:
@jmcgeheeiv - I have played around with sphinx and autodoc and came up with this. Please check if this makes sense at all. 
There is least one problem with this approach: always the latest version of the documentation is shown on GitHub pages, even if it is not released. One way to handle this could be to mention new functionality in the docstrings (e.g. "new  in pyfakefs 2.9"). What do you think?
And, somewhat related (in case I will adapt the docstrings) - what will be the next version? 2.10 or 3.0?
Ah, and I promise to leave this alone for a while when done with the documentation stuff (at least if no bugs come in) :P

jmcgeheeiv:
@mrbean-bremen you needn't promise to do nothing to move pyfakefs forward.

While version 2.10 is correct per semantic versioning, I think we are ready to step up to 3.0: the @mrbean-bremen version with solid Windows support.

So let's follow the standard Python documentation convention of saying, "New in version 3.0."

Where is this documentation rendered?  You might mention the location in README.md.

mrbean-bremen:
Well, I didn't want to bother you too much - you may have more important things to do (and maybe recover from 11/8) ;)
Ok - thanks, I will add the version information in the docstrings where needed. I already added a mention in README.md in the PR - you may have overlooked that one (the URL is http://jmcgeheeiv.github.io/pyfakefs). Note that the documentation is just auto-generated and certainly needs some work, but it may suffice for now.

jmcgeheeiv:
I feel bad for leaving you to your own devices with Sphinx.  I actually use it quite a bit at work.

I think that the Sphnix documentation is more than adequate for now.  Nice job.

mrbean-bremen:
That's completely ok for me - I learned something new (I only used doxygen for documentation generation until now, had no need for Python documentation yet). So I will adapt the docstrings somewhere in the next days and push the PR, if you have no objections.

mrbean-bremen:
I added "New in pyfakefs 3.0" instead of "New in version 3.0" to avoid to confuse it with the python version, so now it may read "New in python 3.3. New in pyfakefs 3.0" or similar. I will merge the PR later, in case you have no further remarks.

jmcgeheeiv:
New, new, new!  I love that word!

mrbean-bremen:
:) If you don't have any further plans for this version, we could indeed go ahead with the 3.0 version.

mrbean-bremen:
@jmcgeheeiv - I'm mostly done with the autodoc stuff, but it doesn't work in the master branch due to my access token not working from your repository (at least this is what I guess, as it works from my branch) - I need this to push the documentation into the gh-pages branch. While this is not a big deal (could always manually commit changes), I would just like t to work... So if you get some time, you could replace the token with one of yours. How to do this:
- remove the current token from .travis.yml (e.g. the *env* stuff)
- install the ruby gem travis (gem install travis) (there is an alternative using node.js and Python which I haven't tried)
- under your GitHub settings, go to Personal Access Tokens, create a new token that has (only) "public_repo" right set and copy it somewhere
- in the repo root of pyfakfs, execute:
travis encrypt GH_TOKEN=<your token> --add env.global
- edit the diff of .travis.yml (the script deletes comments) and commit it

Thanks in advance!

mrbean-bremen:
@jmcgeheeiv - ok, forget my last comment, it may not be worth the trouble. I will just manually update the docs before a release. Speaking of releases - I think it may be time for one more - there have been some features added since 2.9, and at least I am not planning any more changes at the moment.. What do you think?

PR #143, opened 2016-12-04
os.listdir sometime returns a dict instead of a list

MiiRaGe:
Hi,

I found a bug on the os.listdir expected behaviour in python 3.5 (I'm migrating a project from 2.7 to 3.5).

os.listdir on a fake directory returns a dictionnary instead of a list.

I fixed it in the pull request by returning the list of keys if it's a dictionnary. Maybe not the best but I don't know the library enought to dig further.

Kind Regards,
Alexis

### DIFF
--- a/fake_filesystem_test.py
+++ b/fake_filesystem_test.py
@@ -745,7 +745,7 @@ class FakeOsModuleTest(FakeOsModuleTestBase):
         files.sort()
         self.assertEqual(files, sorted(self.os.listdir(directory)))
 
-    def testListdirDictionnary(self):
+    def testListdirReturnsList(self):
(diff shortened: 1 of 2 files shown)


mrbean-bremen:
Nice catch, thank you! I introduced this bug in a recent commit and didn't catch it because the tests always sort the result, which turns the dict into a list.
Can you please slightly change the PR before I merge it:
- in the fix, just use `.keys()` on the result, it is always a dict
- rename the test to something like `testListdirReturnsList`

MiiRaGe:
All done. =)

mrbean-bremen:
Thanks for the contribution!

PR #144, opened 2016-12-15
Increased version to 3.0

mrbean-bremen:


mrbean-bremen:
@jmcgeheeiv - if you are going to push this version to pypi, make sure to pip install pypandoc locally first (needed to convert the README from markdown to rst format, see setup.py).

PR #146, opened 2016-12-20
Fix shutil.move implementation

jannero:
If both the source and destination argument of shutil.move are
directories the source directory is copied under the target directory.
pyfakefs implementation copied the source directory to the target
directory.

The real shutil.move call shutil.copy('/path/to/src', '/path/to/dest')
moves /path/to/src to /path/to/dest/src but the pyfakefs implementation
moved the src directory to /path/to/dest

### DIFF
--- a/pyfakefs/fake_filesystem_shutil.py
+++ b/pyfakefs/fake_filesystem_shutil.py
@@ -285,8 +285,10 @@ class FakeShutilModule(object):
             return dst
 
         source_is_dir = stat.S_ISDIR(self.filesystem.GetObject(src).st_mode)
-        if source_is_dir and self.filesystem.Exists(dst):
-            raise shutil.Error("Destination path '%s' already exists" % dst)
+        if source_is_dir:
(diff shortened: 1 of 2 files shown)


mrbean-bremen:
Thanks a lot for the fix - I will have a look at this in the evening. It looks like the test has also to be adapted.

pyfakefs/fake_filesystem_shutil.py
```diff
@@ -285,7 +285,9 @@ def _destinsrc(src, dst):
             return dst
 
         source_is_dir = stat.S_ISDIR(self.filesystem.GetObject(src).st_mode)
-        if source_is_dir and self.filesystem.Exists(dst):
+        if source_is_dir:
+          dst = os.path.join(dst, os.path.basename(src))
+          if self.filesystem.Exists(dst):
```
mrbean-bremen: Please use `self.filesystem.JoinPaths` instead of `os.path.join` to consider a possibly changed path separator in the fake filesystem.

PR #155, opened 2017-02-02
Fix disk usage calculations when too large file is created

jannero:
Previously if you created file having larger initial contents than the
free space in the fake filesystem, the disk usage counter was decreased
by the size of the initial contents possibly resulting in negative usage.

Basically the issue was that when ChangeDiskUsage is called with size
larger than the amount of free space, space usage is not changed but
an error is raised. Due to an exception, CreateFile removes the created
object that results in new call to ChangeDiskUsage with negative size
of the object.

Previously the object's size was set before ChangeDiskUsage was called
and then when the object was being removed the object's already set size
was used to reduce the filesystem usage (even though that size was never
added to the usage).

### DIFF
--- a/fake_filesystem_test.py
+++ b/fake_filesystem_test.py
@@ -4165,6 +4165,25 @@ class DiskSpaceTest(TestCase):
         self.filesystem = fake_filesystem.FakeFilesystem(path_separator='!', total_size=100)
         self.os = fake_filesystem.FakeOsModule(self.filesystem)
 
+    def testDiskUsageOnFileCreation(self):
+        fake_open = fake_filesystem.FakeFileOpen(self.filesystem)
+
+        total_size = 100
+        self.filesystem.AddMountPoint('mount', total_size)
+        def create_too_large_file():
+            with fake_open('!mount!file', 'w') as dest:
+                dest.write('a' * (total_size + 1))
+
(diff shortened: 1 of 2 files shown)


fake_filesystem_unittest_test.py
```diff
@@ -141,6 +141,22 @@ def test_fakepathlib(self):
...
+
+        self.assertRaises((OSError, IOError), create_too_large_file)
+
+        self.assertEqual(0, self.fs.GetDiskUsage('/mount').used)
+
+        with open('/mount/file', 'w') as dest:
+            dest.write('a' * total_size)
+
+        self.assertEqual(total_size, self.fs.GetDiskUsage('/mount').used)
+
```
mrbean-bremen: Looks nice - thanks! 
Could you please move this test to the class _DiskSpaceTest_ in _fake_filesystem_test.py_ (_fake_filesystem_unitest_test.py_ is used for unit test/patcher related tests) and rename it to conform to the used style (e.g. camel case for test names - seems to be Google's coding style) - I will merge it afterwards.

jannero:
Did the update go unnoticed or is there still something that I should fix?

mrbean-bremen:
@jannero - sorry, for some reason I didn't get notified about your change.

mrbean-bremen:
And thanks for the contribution!

PR #156, opened 2017-02-02
Added pytest plugin support

wesleykendall:
Hello, I love this library and would really like to use it as a pytest fixture in my tests without having to define it myself. This PR adds a pytest plugin as part of the installation for pyfakefs. In other words, installing pyfakefs allows one to do:

```
def my_fake_fs_test(fs):
  fs.CreateFile('/var/data/xx1.txt', contents='Hello!')
  assert os.path.exists('/var/data/xx1.txt')
```

I used your example in conftest.py, along with returning the `fs` attribute of the created `Patcher` object. Everything works so far in my fakefs tests.

Any support for adding this to pyfakefs? I'll gladly write some more docs if desired. Thanks for the feedback


jmcgeheeiv:
"Any support for adding this to pyfakefs?"

You're so funny, @wesleykendall.  Hell yeah there's support for that!

A paragraph or two in README.md would be appropriate for documentation.

This needs a test, which would require pytest, but on the other hand, we need to ensure that pyfakefs works without pytest.  Ideas?

wesleykendall:
@jmcgeheeiv awesome! Here's what I attempted to do in the most recent commits:

1. Updated tox and travis to run py.test on one pytest test file to ensure the fixture is available (had to add a "pip install ." beforehand for the entry point).
2. Removed the conftest.py file. I'm assuming this was just example code. It interfered with my pytest since it also defined the "fs" fixture.
3. Updated the README to have a small subsection about usage of the pytest plugin

Anything else that might be worth adding? I can update the wiki too.

Something worth noting - pytest has deprecated `yield_fixture` for using `fixture`, but using `fixture` will fail with <pytest3 right now. I don't know of an easy way to test out compatibility until pytest officially removes `yield_fixture`.

mrbean-bremen:
You could check for the pytest version to decide which decorator to use. Something along the lines of:
``` python
from distutils.version import LooseVersion

is_pytest2 = LooseVersion(pytest.__version__) < LooseVersion('3')
pytest_fixture = pytest.yield_fixture if is_pytest2 else pytest.fixture

@pytest_fixture
def fs()
...
```

wesleykendall:
@mrbean-bremen @jmcgeheeiv  I just got around this issue with using `addFinalizer` https://github.com/jmcgeheeiv/pyfakefs/pull/156/commits/4deb64411308d97eadba55c80b152d0cd32c1a59

jmcgeheeiv:
@mrbean-bremen, you have been watching this PR more carefully than me, so please take care of the merge as you see fit.  @wesleykendall, thanks a bunch for the contribution!

jmcgeheeiv:
@megapctr, I think you will be interested in this new capability.

mrbean-bremen:
@wesleykendall, thanks a lot!

wesleykendall:
@mrbean-bremen @jmcgeheeiv Thanks! Any idea when this will be released?

mrbean-bremen:
I certainly would plead for a new release here, as this change makes using pytest with pyfakefs much more convenient - though @jmcgeheeiv has to do the release work, and we just had a release...

jmcgeheeiv:
I'll see what I can do this weekend.

jmcgeheeiv:
@wesleykendall, @mrbean-bremen, @megapctr 
pyfakefs version 3.1 released with pytest fixture.

mrbean-bremen:
Thanks - that was fast!

wesleykendall:
Thanks!

PR #158, opened 2017-02-06
Minor changes to copyRealFile

mrbean-bremen:
- do not test the whole string contents as this will not work with other line endings (e.g. on Windows)
- moved setUpClass() code into class as setUpClass() is not available under Python 2.6


mrbean-bremen:
@jmcgeheeiv - please check if this is ok.

jmcgeheeiv:
@mrbean-bremen, your Windows installation of various Python versions is very helpful.

PR #166, opened 2017-03-11
Derive SetLargeFileSizeTest directly from TestCase

jleverenz:
SetLargeFileSizeTest should derive directly from unittest.TestCase instead
of FakeDirectoryUnitTest, to avoid running tests in FakeDirectoryUnitTest
twice.


mrbean-bremen:
Right you are - thanks!

PR #169, opened 2017-03-17
Rename PathManipulationTests to PathManipulationTestBase for consistency

jleverenz:
This base class has no tests, and all other non-concrete test case base classes
in this file use a *TestBase convention.

..small cleanup for your consideration.


mrbean-bremen:
Thanks - cleanup is always good.

Issue #170, opened 2017-03-30
[wanted] Easy way to 'embed' certain directories in the virtual-file system


dwt:
Hi there,

while trying to test my flask application with pyfakefs, I had to do quite some magic to get it working.

My chief problems where, that a) jinja2 (in development mode at least) insists on always loading it's templates from disk and b) babel loads it's datafilee lazily from disk in time for request handling.

Both of these where impossible to workaround by using the additional_skip_modules, so I ended up doing this instead:

```python
    def _restore_accss_to_crucial_files(self):
        from pyfakefs.fake_filesystem_unittest import REAL_OS, REAL_OPEN
        def copy_in_file(real_path):
            real_stat = REAL_OS.stat(real_path)
            with REAL_OPEN(real_path, 'rb') as real_file:
                real_contents = real_file.read()
            fake_file = self.fs.create_file(real_path, st_mode=real_stat.st_mode,
                                           contents=real_contents,
                                           create_missing_dirs=True)
            fake_file.st_ctime = real_stat.st_ctime
            fake_file.st_atime = real_stat.st_atime
            fake_file.st_mtime = real_stat.st_mtime
            fake_file.st_gid = real_stat.st_gid
            fake_file.st_uid = real_stat.st_uid
            return fake_file
        
        def copy_in_directory(directory_path):
            for name in REAL_OS.listdir(directory_path):
                path = REAL_OS.path.join(directory_path, name)
                copy_in_file(path)
            
        template_dir = REAL_OS.path.join(REAL_OS.path.dirname(__file__), 'templates')
        copy_in_directory(template_dir)
        
        import babel
        babel_dir = REAL_OS.path.dirname(babel.__file__)
        copy_in_file(REAL_OS.path.join(babel_dir, 'global.dat'))
        babel_data_dir = REAL_OS.path.join(babel_dir, 'locale-data')
        copy_in_directory(babel_data_dir)
```

This workaround works, but it has the unfortunate side effect of always copying megabytes of data around.

So this is what I'm asking about: Could you provide an easy way to allow access to certain parts of the file system in an easy way, by 'embedding' access to them into the fake filesystem? Bonus points if it's read only...

Some things I can imagine being very useful:
* Allowing (read only) access to site_packages (of the current virtual env?)
* Allow (read only) access to the source code of the current package to load ressources like templates for jinja2

What do you think?

dwt:
Oh, not to mention that hardcoding all those paths is brittle as hell.

[...]

PR #171, opened 2017-04-01
Added methods to access real files

mrbean-bremen:
- allow to add really existing files and directory trees to the fake
file system, with the contents read on demand
- see #170

### DIFF
--- a/pyfakefs/fake_filesystem.py
+++ b/pyfakefs/fake_filesystem.py
@@ -112,7 +112,7 @@ __version__ = '3.2'
 
 PERM_READ = 0o400  # Read permission bit.
 PERM_WRITE = 0o200  # Write permission bit.
-PERM_EXE = 0o100  # Write permission bit.
+PERM_EXE = 0o100  # Execute permission bit.
 PERM_DEF = 0o777  # Default permission bits.
 PERM_DEF_FILE = 0o666  # Default permission bits (regular file)
 PERM_ALL = 0o7777  # All permission bits.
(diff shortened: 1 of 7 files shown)


mrbean-bremen:
@jmcgeheeiv - please review!

pyfakefs/fake_filesystem.py
```diff
@@ -1003,7 +1014,7 @@ def SplitDrive(self, path):
                 # UNC path handling is here since Python 2.7.8, back-ported from Python 3
                 if sys.version_info >= (2, 7, 8):
                     if (path[0:2] == self.path_separator * 2) and (
-                            path[2:3] != self.path_separator):
+                                path[2:3] != self.path_separator):
```
mrbean-bremen: Didn't notice these changes - PyCharm has sometimes its own ideas about hanging indents, obviously I did an auto-format accidentally. I will revert this.

mrbean-bremen:
I added an example as requested. I had to use `REAL_OPEN` to compare the content with the real file, but you won't need this in a real test, so I think the API can be used in unit tests without problems.

mrbean-bremen:
Also - as `CopyRealFile` has at least one additional feature  (copy to another location), I think it has its own value,  but I don't think we need another addition to TestCase - the methods in fake_filesystem should be sufficient.

mrbean-bremen:
@dwt had similar problems as you mentioned, and after reproducing them (or something related ) I found that the call to `os.path.exists` that is made from the original  `os.listdir`, was the problem in this case. I'm not really sure if this is a good fix for that.

mrbean-bremen:
I may have another go at this later and try to reproduce the problem in a unit test.

mrbean-bremen:
@jmcgeheeiv - I added a test that reproduces the problem. please review!
Note that I rebased against master.

mrbean-bremen:
As to the explanation why the problem happened in the first place: 
Other than the other patched modules, `path` is actually implemented in two modules: `ntpath`/`posixpath` (one of which is imported as `path` by `os`) and `genericpath`, which implements stuff common for both systems. If calling `os.path.exists` from inside the unit test, the real `os` module is used as it should be, but it calls the implementation in `genericpath`, which calls `os.stat` on its part. If `genericpath` is not added to SKIPNAMES, this last `os` will be stubbed out, so `stat` is called on the fake file system.

jmcgeheeiv:
Oh, dear.  Could you resolve this conflict and merge it?

Well done, @mrbean-bremen.

mrbean-bremen:
Done - caused that conflict myself. Thanks!

PR #175, opened 2017-04-05
Patch only modules with fake_filesystem_unittest.Patcher

jmcgeheeiv:
One of my modules has a `path` attribute, and I was dismayed to find that it got patched.

This branch takes extra care to patch only modules.

Code review please, @mrbean-bremen.


mrbean-bremen:
I corrected the minor typo myself and amended your commit (hope you don't mind that), before I merged it. And I used "Rebase and merge" just to find out how this works - never used it before...

PR #176, opened 2017-04-05
copyRealFile fake_file_path defaults to real_file_path

menge101:
Changed TestCase.copyRealFile to default the value of fake_file_path to the same value as real_file_path, when fake_file_path is not provided/None.

### DIFF
--- a/fake_filesystem_unittest_test.py
+++ b/fake_filesystem_unittest_test.py
@@ -207,6 +207,29 @@ class TestCopyRealFile(TestPyfakefsUnittestBase):
             self.copyRealFile(real_file_path, fake_file_path,
                               create_missing_dirs=False)
 
+    def testCopyRealFileNoDestination(self):
+        real_file_path = __file__
+        fake_file = self.copyRealFile(real_file_path)
+
+        self.assertTrue('class TestCopyRealFile(TestPyfakefsUnittestBase)' in self.real_string_contents,
+                        'Verify real file string contents')
+        self.assertTrue(b'class TestCopyRealFile(TestPyfakefsUnittestBase)' in self.real_byte_contents,
+                        'Verify real file byte contents')
+
+        # note that real_string_contents may differ to fake_file.contents due to newline conversions in open()
+        self.assertEqual(fake_file.byte_contents, self.real_byte_contents)
+
+        self.assertEqual(fake_file.st_mode, self.real_stat.st_mode)
+        self.assertEqual(fake_file.st_size, self.real_stat.st_size)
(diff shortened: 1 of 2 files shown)


pyfakefs/fake_filesystem_unittest.py
```diff
@@ -152,6 +152,8 @@ def copyRealFile(self, real_file_path, fake_file_path=None,
                   offers the option to enable atime, and older versions of \
                   Linux may also modify atime.
         """
+        if not fake_file_path:
```
jmcgeheeiv: if fake_file_path is not None:

pyfakefs/fake_filesystem_unittest.py
```diff
@@ -130,7 +130,7 @@ def copyRealFile(self, real_file_path, fake_file_path=None,
 
         Args:
           real_file_path: Path to the source file in the real file system.
-          fake_file_path: path to the destination file in the fake file system.
+          fake_file_path: path to the destination file in the fake file system.  Defaults to real_file_path.
```
jmcgeheeiv: Please indent.

fake_filesystem_unittest_test.py
```diff
@@ -207,6 +207,29 @@ def testCopyRealFile(self):
             self.copyRealFile(real_file_path, fake_file_path,
                               create_missing_dirs=False)
 
+    def testCopyRealFileNoDestination(self):
```
jmcgeheeiv: Here you are testing happens when you do not specify the destination path.  It is adequate to test only for the existence of the destination file.

The previous test tested all the details like the contents and stat.

menge101: Oh yeah, i Just copied the existing test and changed the parameters on the method call.

OK, I'll trim all that out.

pyfakefs/fake_filesystem_unittest.py
menge101: Do you mean:
   `if fake_file_path is None`   ? 
`if fake_file_path is not None` would be the case where it already has a set value.

pyfakefs/fake_filesystem_unittest.py
menge101: Sure:

Like this?

fake_file_path: path to the destination file in the fake file system.
            Defaults to real_file_path.

Or this?

fake_file_path: path to the destination file in the fake file system.
                         Defaults to real_file_path.

?

jmcgeheeiv:
Well done.  Thank you and welcome to the community of pyfakefs contributors!

menge101:
Thanks, love the library, happy to contribute.

PR #184, opened 2017-05-14
Added support for errors argument in FakeFileOpen

mrbean-bremen:


mrbean-bremen:
While fixing an obscure encoding problem I noticed that the `errors` argument in `io.open` argument is not supported, As this was an easy one, I decided to add it before the relase...

PR #185, opened 2017-05-16
Added lazy evaluation of real directories

mrbean-bremen:
- real directory contents are only added to the fake file system on
demand (makes adding large directory trees faster)
- refactored real file/directory access to use own classes


mrbean-bremen:
I use tox locally, which creates the virtual environments for all versions in a .tox directory under pyfakefs. This led to one of the unit tests (real directory tree) to run about 10 seconds for each python version, as it enumerated all files in these directories recursively. So I decided to add lazy evaluation also for real directory contents...
This wil be my last change for now (if no bug comes in), so after that you can go ahead with the release if you want.

jmcgeheeiv:
Wow.  Big job.

PR #212, opened 2017-06-16
Added possibility to use open file descriptor as path

### LINKED ISSUE #205: Add support for open file descriptor as path argument
Since Python 3.3, a number of `os` functions support using an open file descriptor as path argument.
These include `os.utime`, `os.chmod`, `os.chdir`, 'os.chown', `os.listdir`, `os.stat` and `os.lstat`.
The value of `os.supports_fd` shall be considered in the implementation.

Needed in #194 for `os.listdir`.

mrbean-bremen:
- added for os.utime, os.chmod, os.chdir, os.chown, os.listdir, os.stat and os.lstat
- fixes #205

### DIFF
--- a/CHANGES.md
+++ b/CHANGES.md
@@ -4,6 +4,9 @@ The release versions are PyPi releases.
 ## Version 3.3 (as yet unreleased)
 
 #### New Features
+  * Added support for open file descriptor as path argument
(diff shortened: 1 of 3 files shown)


mrbean-bremen:
@jmcgeheeiv - I'm going to merge this stuff without review. Rght now we seem to be getting a flood of new bugs from the TSTL tool, and I want to have these things out of the way. We can always refactor afterwards...

PR #328, opened 2017-11-13
Removed Python 2.6 support

### LINKED ISSUE #293: Remove support for Python 2.6
With Python 2.6 being at end of life now almost 4 years, we may drop the support IMHO. 
Most other packages are also dropping support, so there is not much sense to keep it. 
This will also allow to to cleanup the code by removing the Python 2.6 specific workarounds.
I propose to do this for the 3.4 release, alongside with the PEP-8 changes.

mrbean-bremen:
- closes #293

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -18,8 +18,6 @@ language: python
 
 matrix:
     include:
-        - python: 2.6
-          env: PYTHON=py26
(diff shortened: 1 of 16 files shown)


mrbean-bremen:
oops, duplicate

PR #333, opened 2017-12-09
Added support to fake out the backported scandir

### LINKED ISSUE #332: Pyfakefs doesn't work with Python 2.6+ backport of os.scandir
pyfakefs doesn't work with [the backport](https://github.com/benhoyt/scandir) of [`os.scandir`](https://docs.python.org/3.6/library/os.html#os.scandir) Is this possible or does it have the same [limitations as lxml](https://github.com/jmcgeheeiv/pyfakefs#compatibility)?

Currently the backport, just like in #119, doesn't find any directories created by pyfakefs and raises an OSError:

```
OSError: [Errno 2] No such file or directory: 'root'
```

mrbean-bremen:
- refactored Patcher for easier possibility to add new modules
- moved fake scandir implementation into separate module and use it both for os.scandir and scandir.scandir
- shall fix #332

### DIFF
--- a/fake_filesystem_test.py
+++ b/fake_filesystem_test.py
@@ -30,7 +30,7 @@ import unittest
 
 from pyfakefs import fake_filesystem
 from pyfakefs.fake_filesystem import FakeFileOpen
-
+from pyfakefs.fake_filesystem_unittest import has_scandir
 
 class _DummyTime(object):
     """Mock replacement for time.time. Increases returned time on access."""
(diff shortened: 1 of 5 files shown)


mrbean-bremen:
A few notes:
- I did not add a generic mechanism to add a new module to `Patcher`, but doing this should be easy now
- the tests for `scandir.scandir()` are only running if `scandir` is installed - I didn't do this in the Travis build, only tested it locally
- this PR will clash with the other PR, so one of them will have to be rebased

PR #348, opened 2018-02-20
Test case mixin

jannero:
This PR splits TestCase into TestCaseMixin and TestCase. The TestCaseMixin class can be used with other than unittest.TestCase TestCase classes (for example, django.test.TestCase or asynctest.TestCase)

### DIFF
--- a/pyfakefs/fake_filesystem_unittest.py
+++ b/pyfakefs/fake_filesystem_unittest.py
@@ -103,7 +103,92 @@ def load_doctests(loader, tests, ignore, module,
     return tests
 
 
-class TestCase(unittest.TestCase):
+class TestCaseMixin(object):
+    """Test case mixin that automatically replaces file-system related
+    modules by fake implementations.
+
+    Attributes:
+        additional_skip_names: names of modules inside of which no module
+            replacement shall be performed, in addition to the names in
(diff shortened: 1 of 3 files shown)


pyfakefs/fake_filesystem_unittest.py
```diff
@@ -103,7 +103,92 @@ def load_doctests(loader, tests, ignore, module,
...
+                                use_dynamic_patch=self.use_dynamic_patch,
+                                modules_to_reload=self.modules_to_reload,
+                                modules_to_patch=self.modules_to_patch)
+
+        self._stubber.setUp()
+        self.addCleanup(self._stubber.tearDown)
+
+
+class TestCase(TestCaseMixin):
     """Test case class that automatically replaces file-system related
```
mrbean-bremen: Thanks for contributing this!
One question though - if you derive TestCase only from the mix-in, and not from `unittest.TestCase` as before, will that not break existing tests? I would have expected:
``` python
class TestCase(unittest.TestCase, TestCaseMixin):
    ...
```

jannero: Whoops. Good that there is riviews. I had intention to do that but missed it.

mrbean-bremen: Thanks - looks good now!

PR #349, opened 2018-02-22
Add flake8 to testing

cclauss:
* FileNotFoundError, FileExistsError, NotADirectoryError exist in Python 3 but [not in Python 2](https://travis-ci.org/jmcgeheeiv/pyfakefs/jobs/344658535#L479-L512).
* file(), long, unicode exist in Python 2 but [not in Python 3](https://travis-ci.org/jmcgeheeiv/pyfakefs/jobs/344658538#L475-L502).

### DIFF
--- a/.travis/run_tests.sh
+++ b/.travis/run_tests.sh
@@ -5,4 +5,10 @@ if [[ $TRAVIS_OS_NAME == 'osx' ]]; then
 fi
 
 python --version
+pip --version
+pip install flake8
+# stop the build if there are Python syntax errors or undefined names
+flake8 . --count --select=E901,E999,F821,F822,F823 --show-source --statistics
+# exit-zero treats all errors as warnings.  The GitHub editor is 127 chars wide
+flake8 . --count --exit-zero --max-complexity=10 --max-line-length=127 --statistics
 python -m tests.all_tests


mrbean-bremen:
Thanks for the contribution! I had planned to add pylint sometime (there is still an old pylint.rc lying around in the root dir), but never got to it.
I haven't used flake much (more used to pep8 and pylint) - are there any options to disable warnings in the code at specific places like for pylint? I don't want to have these warnings about `long` etc there, as they are expected, and in some places there are already pylint directives in the code.
Also, it would be nice to exclude the `pathlib` stuff from Python 2 builds, as it is Python 3 only, to exclude these warnings too.
As for line length - I personally am used to something like 120 characters per line, but lately tried to change it to the standard 79 characters, as this is enforced in another project I'm working on. 

So, basically, I like the PR, but to be more helpful it would make sense to exclude the unwanted warnings and errors and fix the rest (otherwise it will just be ignored). WDYT?

cclauss:
* Pylint does not find undefined names.
* Put `# noqa` on lines you want ignored.
* —exclude directories or individual files

mrbean-bremen:
Ok, thanks.
That pylint does not find undefined names is strange - it should. 
The rest seems to be like in pep8 - I still like pylint better because you have better control - #noqa disables checks for the line at all.
I may have another look in the evening - can't promise, though, after that, I will have access to a computer only somewhere next week.

cclauss:
flake8 is PEP8 (now called pycodestyle) + pyflakes + McCabe cyclomatic complexity so it is a complete superset of PEP8.  All of these tools (including pylint) live at https://github.com/pycqa

cclauss:
You can also write `# noqa: F821` to only suppress one particular alert

mrbean-bremen:
Ah, ok - thanks, didn't know that. 
I will check if I can get this to build without errors first, remove the --exit-zero flag, and merge it back afterwards. As I said, I will have time for this not before next week, but thanks for the reminder - something that is on my ToDo list for some time now...

mrbean-bremen:
Ok, I will merge now and handle the problems later - thanks for the contribution!

PR #351, opened 2018-02-23
Fix undefined names in Python3

cclauss:
The _try: except NameError:__ approach follows the Python porting best practice [Use feature detection instead of version detection](https://docs.python.org/3/howto/pyporting.html#use-feature-detection-instead-of-version-detection) and keeps the linters at bay.

The alternative approach uses the very popular [six](http://six.readthedocs.io) module instead.

### DIFF
--- a/tests/fake_filesystem_unittest_test.py
+++ b/tests/fake_filesystem_unittest_test.py
@@ -57,10 +57,10 @@ class TestPyfakefsUnittest(TestPyfakefsUnittestBase):  # pylint: disable=R0904
     def test_file(self):
         """Fake `file()` function is bound"""
         self.assertFalse(os.path.exists('/fake_file.txt'))
-        with file('/fake_file.txt', 'w') as f:
+        with file('/fake_file.txt', 'w') as f:  # noqa: F821 is only run on Py2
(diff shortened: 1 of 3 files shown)


mrbean-bremen:
Thanks again! I will have a closer look once I I'm back, somewhere next week.

tests/fake_filesystem_unittest_test.py
```diff
@@ -57,10 +57,10 @@ class TestPyfakefsUnittest(TestPyfakefsUnittestBase):  # pylint: disable=R0904
     def test_file(self):
         """Fake `file()` function is bound"""
         self.assertFalse(os.path.exists('/fake_file.txt'))
-        with file('/fake_file.txt', 'w') as f:
+        with open('/fake_file.txt', 'w') as f:
             f.write("This test file was created using the file() function.\n")
         self.assertTrue(self.fs.exists('/fake_file.txt'))
-        with file('/fake_file.txt') as f:
+        with open('/fake_file.txt') as f:
             content = f.read()
```
mrbean-bremen: These are actually wanted - it explicitely tests that `file` for Python2 id faked.
The rest looks good!

PR #361, opened 2018-04-06
Include tests in PyPI tarball

dotlambda:


mrbean-bremen:
Ok, thanks - makes sense.

PR #379, opened 2018-04-20
Remove unnecessary shebang line

kbabioch:
This removes an unnecessary shebang line from `fake_filesystem_shutil.py`, as
this file is a Python module and isn't invoked via CLI.


mrbean-bremen:
Thanks!

Issue #408, opened 2018-05-30
Support pathlib2


sk-:
First of all thanks for this project, it's really useful. However, it'd be great if there was support for [`pathlib2`](https://github.com/mcmtroffaes/pathlib2).

In my code I'm importing it as `import pathlib2 as pathlib`. Which gets patched by `pyfakefs`, however the patch uses the system `pathlib`, which means that in my case it only works for Python 3.5 and 3.6, as I'm using the `exist_ok` argument of `mkdir`. In Python 3.4 I get the following error:

```
TypeError: mkdir() got an unexpected keyword argument 'exist_ok' 
```

And in Python 2.7 the module is not patched at all.

I already managed to get this to work in Python 2.7 (and probably in 3.4),

In my test I had to add

```py
def setUp():
  from pyfakefs import fake_pathlib
  self.modules_to_patch = {'pathlib': fake_pathlib.FakePathlibModule}
  self.setUpPyfakefs()
```

and I had to change `fake_pathlib.py` as following
```py
try:
  import pathlib2 as pathlib
except ImportError:
  import pathlib
try:
  from urllib.parse import quote_from_bytes as urlquote_from_bytes
except ImportError:
  from urllib import quote as urlquote_from_bytes
```

Let me know if you would be interested in accepting this change, so I could prepare a patch.

mrbean-bremen:
Sure, that sounds helpful. I actually wasn't aware of `pathlib2`, but we did something similar with the `scandir` module which is a backport of 'scandir` to Python < 3.5.
And nice to know that the `modules_to_patch` has been used by someone :)

matthew16550:
There is another difficulty using `pathlib2` with Python 2.7, the version checking in `FakePath.resolve()` prevents using `strict=False`.

e.g.
```python
def test_resolve(fs):
    Path('foo').resolve(False)
```

Causes:
 `TypeError: resolve() got an unexpected keyword argument 'strict'`

mrbean-bremen:
Yes, the version checking has been made for `pathlib` only - the checks have to be changed for `pathlib2` support.
@sk- - are you still working on the PR? Do you need help?

sk-:
@mrbean-bremen sorry I don't have the time to prepare the PR. Feel free to use any of the code snippets if they are of any help.

matthew16550:
@mrbean-bremen I probably have time this weekend if you would like me to make a PR?

mrbean-bremen:
[...]

PR #422, opened 2018-08-07
Add support for pathlib2 (#408)

matthew16550:


pyfakefs/tests/all_tests_without_extra_packages.py
```diff
@@ -0,0 +1,19 @@
+import sys
```
matthew16550: I'm not very happy with the approach in this file but don't have a better idea.

pyfakefs/fake_pathlib.py
```diff
@@ -496,10 +504,10 @@ def open(self, mode='r', buffering=-1, encoding=None,
         """
         if self._closed:
             self._raise_closed()
-        return FakeFileOpen(self.filesystem)(
+        return FakeFileOpen(self.filesystem, use_io=True)(
```
matthew16550: I added `use_io=True` 3 times in this file to make some py27 tests pass on Windows.  Not sure if it's the best approach, maybe I should add something to the `self._use_io` logic in `FakeFileOpen.__init__()`  instead?

The test failures can be seen in [AppVeyor](https://ci.appveyor.com/project/jmcgeheeiv/pyfakefs/build/3.4.172/job/3pbw1m03nw1swqg2), e.g:
```
  File "pyfakefs\tests\fake_filesystem_unittest_test.py", line 131, in test_fakepathlib
  File "pyfakefs\fake_pathlib.py", line 508, in open
  File "pyfakefs\fake_filesystem.py", line 4765, in __call__
TypeError: _call_ver2() takes at most 6 arguments (7 given)
```
```
  File "pyfakefs\tests\fake_pathlib_test.py", line 508, in test_read_text
    self.assertEqual(file_path.read_text(), 'foo')
  File "pyfakefs\fake_pathlib.py", line 527, in read_text
    errors=errors) as f:
  File "pyfakefs\fake_filesystem.py", line 4765, in __call__
    return self._call_ver2(*args, **kwargs)
TypeError: _call_ver2() got an unexpected keyword argument 'errors'
```

pyfakefs/tests/fake_pathlib_test.py
```diff
@@ -382,35 +413,35 @@ def test_resolve(self):
             self.path(
                 self.os.path.realpath(self.make_path('antoine', 'setup.py'))))
 
-    @unittest.skipIf(sys.version_info >= (3, 6),
-                     'Changed behavior in Python 3.6')
     def test_resolve_nonexisting_file(self):
+        before_pathlib_36()
         path = self.path('/foo/bar')
-        self.assertRaises(FileNotFoundError, path.resolve)
```
jmcgeheeiv: I like these little cleanups.  Well done.

pyfakefs/tests/fake_pathlib_test.py
```diff
@@ -21,24 +21,55 @@
...
+from pyfakefs.extra_packages import pathlib, pathlib2
+from pyfakefs.helpers import text_type
+from pyfakefs.tests.test_utils import RealFsTestCase, TestCase
 
 is_windows = sys.platform == 'win32'
 
 
+def needs_pathlib_35():
+    if sys.version_info < (3, 5) and not pathlib2:
+        raise unittest.SkipTest('New in version 3.5')
```
mrbean-bremen: Nice - though I would rename these to be consistent with the other skip functions (e.g. either `skip_if_xxx` or `check_if_xxx`).

pyfakefs/tests/fake_pathlib_test.py
```diff
@@ -467,35 +498,53 @@ def test_exists(self):
 
     def test_open(self):
         self.create_dir(self.make_path('foo'))
-        self.assertRaises(OSError,
+        self.assertRaises(IOError if TestCase.is_python2 else OSError,
```
mrbean-bremen: `IOError` is an alias for `OSError` in Python 3, so you may just check for `IOError`.

requirements.txt
```diff
@@ -1 +1,3 @@
+pathlib2>=2.3.2
 pytest>=2.8.6
+scandir>=1.8
```
mrbean-bremen: I would prefer to leave this file as is, as it contains the requirements needed for `pyfakefs`, and put the extra requirements into an extra requirements file.

pyfakefs/tests/all_tests_without_extra_packages.py
matthew16550: Unfortunately `pytest` depends on `pathlib2` which depends on `scandir`!

matthew16550: Aha I see what's happening, their [setup.py](https://github.com/pytest-dev/pytest/blob/master/setup.py) adds `pathlib2` as a dependency when Python < 3.6 so if we find no other approach I'll update the comment in `extra_requirements.txt`.

mrbean-bremen: Ah, I see - well, I can't think of another approach right now in this case. Maybe @jmcgeheeiv has an idea, otherwise I would it leave as is.

matthew16550:
Yep that fixed `TestCopyOrAddRealFile`thanks.

I've made all the changes you suggested so the only thing left is a final decision on `all_tests_without_extra_packages.py`.

mrbean-bremen:
Looks fine to me - thanks a lot! @jmcgeheeiv will have a seconf review and handle the merge.

PR #434, opened 2018-09-26
Changed patch behavior to rely on module names instead of imported names

mrbean-bremen:
- adapted tests and documentation


mrbean-bremen:
@jmcgeheeiv - please check if you see any potential problems with this approach.
The performance is slighly worse because of the added dict lookup, but still acceptable in my opinion.

PR #439, opened 2018-10-07
Adapted file timestamp update to real fs behavior

### LINKED ISSUE #435: st_ctime/st_atime/st_mtime differences in behaviour compared to linux
Hi, I might have found a problem depending on the expected behaviour.
I noticed that the times st_ctime/st_atime/st_mtime are behaving differently in regard to when the times are updated and if they are updated.
I created two testfiles that run the same sequences on the real fs and on pyfakefs and checks it, see travis build:
https://travis-ci.org/simonfagerholm/pyfakefs/builds/434572170
The test files are fake_stat_time_test.py and real_stat_time_test.py

On linux the real_stat_time_test.py passes all tests (confirming that the expected result corresponds to linux fs) but many fail on pyfakefs. However, on mac os the real_stat_time_test.py also fails a few times so linux and mac behaves differently.

How is pyfakefs supposed to work?

mrbean-bremen:
- tests originally by @simonfagerholm, adapted
- fixes #435

### DIFF
--- a/CHANGES.md
+++ b/CHANGES.md
@@ -18,6 +18,8 @@ keep using pyfakefs 3.4.3, or upgrade to a newer Python version.
 #### Infrastructure
 
 #### Fixes
+  * file timestamps are now updated more according to the real behavior
+    ([#435](../../issues/435))
 
 ## [Version 3.4.3](https://pypi.python.org/pypi/pyfakefs/3.4.3)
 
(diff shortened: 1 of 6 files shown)


mrbean-bremen:
@simonfagerholm, @jmcgeheeiv 
This fixes the tests - please have a look.
Note, that I did not add the timestamp resolution stuff, will do this separately later.

mrbean-bremen:
I disabled the tests for MacOS to avoid the long test times - we can always enable them if needed.

PR #441, opened 2018-10-10
Fix pytest when pyfakefs + future is installed

asottile:
`python-future` is notorious for breaking modules which use `try:` / `except:`
to import modules based on version.  In this case, `pyfakefs` imported the
backported `builtins` module which changes the semantics of the `open()`
function.  `pyfakefs` then monkeypatches `linecache` which breaks any module
which attempts to use `linecache` (in this case `pytest`).

The downstream issue is https://github.com/pytest-dev/pytest/pull/4074

### DIFF
--- a/pyfakefs/pytest_plugin.py
+++ b/pyfakefs/pytest_plugin.py
@@ -10,15 +10,16 @@ def my_fakefs_test(fs):
 """
 
 import linecache
+import sys
 
 import py
 import pytest
 
 from pyfakefs.fake_filesystem_unittest import Patcher
 
-try:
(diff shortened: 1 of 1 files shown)


mrbean-bremen:
Thanks - wouldn't have expected that! Is this something that could cause problems in other places? We use try/except for import in other places in pyfakefs, too.

asottile:
I did an audit of the other imports while implementing this and you should be good (they were (essentially) all for `pathlib` / `scandir` which `future` doesn't "backport")

mrbean-bremen:
Ok, thank you!

Issue #448, opened 2018-11-06
Support fakefs "pause and resume"


majuscule:
Feature request to support "pausing" and "resuming" use of the fakefs, allowing for a test process to run code that the fakefs does not support, then adding new data to the fakefs from the real fs upon resumption.

mrbean-bremen:
@majuscule - please check if this works for you.

mrbean-bremen:
Functionality is implemented in the released version - closing.

majuscule:
Looks perfect! Thank you. I don't actually see a release out though? `--pre` or otherwise. Also while I'm here I'll add one more tiny idea - it'd be nice if this was supported as a context manager to be able to do:

```python
with fs.pause():
     # my syscalls
```

mrbean-bremen:
This should be in the latest release (3.5.3), and had already been added in 3.5.2.
About the context manager - I have been thinking about this, but I wasn't sure where to add it (Patcher, UnitTest, or fake_filesystem). I can add this later...

majuscule:
Whoops, my mistake. Thanks!

PR #451, opened 2018-11-08
Added support for pause/resume patching of the file system modules

mrbean-bremen:
- can be called from the unit test, the patcher or the fs
- see #448

### DIFF
--- a/pyfakefs/tests/pytest/pytest_plugin_test.py
+++ b/pyfakefs/tests/pytest/pytest_plugin_test.py
@@ -1,7 +1,24 @@
 """Tests that the pytest plugin properly provides the "fs" fixture"""
 import os
+import tempfile
 
 
 def test_fs_fixture(fs):
     fs.create_file('/var/data/xx1.txt')
     assert os.path.exists('/var/data/xx1.txt')
+
+
+def test_pause_resume(fs):
+    fake_temp_file = tempfile.NamedTemporaryFile()
+    assert fs.exists(fake_temp_file.name)
+    assert os.path.exists(fake_temp_file.name)
+    fs.pause()
+    assert fs.exists(fake_temp_file.name)
+    assert not os.path.exists(fake_temp_file.name)
+    real_temp_file = tempfile.NamedTemporaryFile()
+    assert not fs.exists(real_temp_file.name)
+    assert os.path.exists(real_temp_file.name)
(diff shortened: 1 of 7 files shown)


mrbean-bremen:
@jmcgeheeiv - please have a look!

pyfakefs/tests/pytest/pytest_plugin_test.py
```diff
@@ -1,7 +1,24 @@
...
+    fs.pause()
+    assert fs.exists(fake_temp_file.name)
+    assert not os.path.exists(fake_temp_file.name)
+    real_temp_file = tempfile.NamedTemporaryFile()
+    assert not fs.exists(real_temp_file.name)
+    assert os.path.exists(real_temp_file.name)
+    fs.resume()
+    assert not os.path.exists(real_temp_file.name)
+    assert os.path.exists(fake_temp_file.name)
+
```
jmcgeheeiv: Do `resume()` again, expect an exception.

pyfakefs/tests/pytest/pytest_plugin_test.py
```diff
@@ -1,7 +1,24 @@
...
+def test_pause_resume(fs):
+    fake_temp_file = tempfile.NamedTemporaryFile()
+    assert fs.exists(fake_temp_file.name)
+    assert os.path.exists(fake_temp_file.name)
+    fs.pause()
+    assert fs.exists(fake_temp_file.name)
+    assert not os.path.exists(fake_temp_file.name)
+    real_temp_file = tempfile.NamedTemporaryFile()
+    assert not fs.exists(real_temp_file.name)
+    assert os.path.exists(real_temp_file.name)
```
jmcgeheeiv: Do `pause()` again, expect an exception.

pyfakefs/tests/pytest/pytest_plugin_test.py
jmcgeheeiv: Actually I strongly agree.  The state before the invocation is irrelevant as long as it's paused after `pause()` and resumed after `resume()`.  The API should always be permissive in this way.

mrbean-bremen: Ok, glad that we are on the same page - though it took me an iteration to come to that.
Are there any changes you want to see?

PR #455, opened 2018-11-28
Set initial user and group IDs to real IDs

mrbean-bremen:
- added tests running as root to test both root and non-root cases
- adapted code and tests to consider root access

### DIFF
--- a/.travis/run_tests.sh
+++ b/.travis/run_tests.sh
@@ -5,19 +5,27 @@ if [[ $TRAVIS_OS_NAME == 'osx' ]]; then
 fi
 
 if ! [[ $VM == 'Docker' ]]; then
-python --version
-pip --version
-pip install flake8
+    python --version
+    pip --version
+    pip install flake8
 
-EXCLUDES='--exclude get-pip.py'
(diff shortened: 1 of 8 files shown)


mrbean-bremen:
I have one test failing in Python 3.7. I tried to skip it, but for some reason that does not work with root user.

jmcgeheeiv:
Instead of using sudo, we can use my Docker container method to run the test as root without endangering the host.

mrbean-bremen:
> Instead of using sudo, we can use my Docker container method to run the test as root without endangering the host.

The problem with the Docker image is that it runs only one Python version - or we have to configure it accordingly for each version. That is actually what Travis.CI does for us, and as it runs in cloned VMs that are trashed after the tests, there is no problem with endangering the host (otherwise they wouldn't provide sudo without password...).
Though as I wrote elsewhere, it would be a good idea also to run the tests in the provided Docker image to make sure that it works correctly.

My problem was that I got the wrong environment if run with sudo, so the system python (Python 2.7) was used instead of the Python version installed in the virtual environment. Inheriting PATH fixed this.

mrbean-bremen:
Merging after resolviong conflicts again...

PR #456, opened 2018-11-29
Execute the tests in a Docker container

jmcgeheeiv:
I got my Dockerfile set up with the right locale and with non-root user `pyfakefs`.

The main purpose of this test is an example Dockerfile.  The tests are not actually run in Docker.

### DIFF
--- a/.dockerignore
+++ b/.dockerignore
@@ -1,0 +1,3 @@
+docs
+.git
+.github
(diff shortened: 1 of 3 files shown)


mrbean-bremen:
Good idea to provide this. Maybe can also add one test that actually uses it - just to make sure that it runs in a Docker image, and on the latest Ubuntu (Travis uses 14.04).
Though I still would add the tests running as root using sudo in the Travis VM, so we test it for each Python version (running the tests is fast, anyway). Merging...

PR #457, opened 2018-11-30
Added running tests in Docker image

mrbean-bremen:
- run for Python 3.6 Linux build only

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -43,6 +43,8 @@ matrix:
         - os: osx
           language: generic
           env: PYTHON=py37
+        - language: minimal
+          env: VM=Docker
 
 install:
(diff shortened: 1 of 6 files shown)


mrbean-bremen:
Note that I have skipped the one failing real FS test if inside a Docker container, though it also might be related to the Ubuntu version.
I'm also not too happy to run this together with the Python 3.6 build - I would prefer a separate build, but couldn't get it to mark alongside the matrix configuration in Travis.CI.
Maybe I will add a new matrix entry instead...

mrbean-bremen:
Ok, I moved the Docker build into an extra matrix entry, this is better now. I will merge this now.

PR #471, opened 2019-03-13
FakeIOModule.open argument renaming: file_name to file

kmerenkov:
Changed argument name in `FakeIOModule.open` from `file_name` to `file`.
It makes pyfakefs compatible with the library atomicwrites that uses `kwargs` to pass arguments to `io.open` function.
See https://github.com/untitaker/python-atomicwrites/blob/890e7aed4a1fd3b3724dc64503757617996c3978/atomicwrites/__init__.py#L180

In python documentation argument is also named `file`:
py3: https://docs.python.org/3/library/functions.html#open
py2: https://docs.python.org/2.7/library/io.html#io.open

### DIFF
--- a/pyfakefs/fake_filesystem.py
+++ b/pyfakefs/fake_filesystem.py
@@ -4553,7 +4553,7 @@ class FakeIoModule(object):
         self.filesystem = filesystem
         self._io_module = io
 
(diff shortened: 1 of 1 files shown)


mrbean-bremen:
Thanks - this makes sense!

PR #477, opened 2019-04-26
document existence of temporary directories

### LINKED ISSUE #476: Document Temp Dir Creation
**Is your feature request related to a problem? 
I recently ran into the the issue desribed in #329, this caused me to waste a significant amount of time trying to figure out why my tests were passing on my Mac but not on my build system running Ubuntu. I assumed the file system was blank and was creating the `/tmp` directory in my tests and of course this worked fine locally because the temp directory pyfakefs creates for MacOS is `/var`.

**Describe the solution you'd like**
Document that the filesystem is not blank and explain the difference between MacOs and Linux.

**Describe alternatives you've considered**
n/a

I'm willing to make the PR to if someone can tell me where this note should live.

svozza:
Closes #476. I just kept it short and sweet but just let me know if there's anything you want me to change.

### DIFF
--- a/docs/usage.rst
+++ b/docs/usage.rst
@@ -468,7 +468,8 @@ OS temporary directories
 ~~~~~~~~~~~~~~~~~~~~~~~~
 
 As ``pyfakefs`` does not fake the ``tempfile`` module this means that a
-temporary directory is required to ensure ``tempfile`` works correctly.
-This means that any newly created fake file system will always have
-either a directory named ``/tmp`` when running on Linux or Unix systems
-or ``/var`` when running on MacOs.
+temporary directory is required to ensure ``tempfile`` works correctly, e.g.,
+``tempfile.gettempdir()`` will return a valid value. This means that any
+newly created fake file system will always have either a directory named
+``/tmp`` when running on Linux or Unix systems, ``/var/folders/<hash>/T``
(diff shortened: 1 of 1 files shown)


docs/usage.rst
```diff
@@ -460,3 +460,15 @@ Here is the same code using a context manager:
...
+---------------
+
+OS temporary directories
+~~~~~~~~~~~~~~~~~~~~~~~~
+
+As ``pyfakefs`` does not fake the ``tempfile`` module this means that a
+temporary directory is required to ensure ``tempfile`` works correctly.
+This means that any newly created fake file system will always have
+either a directory named ``/tmp`` when running on Linux or Unix systems
+or ``/var`` when running on MacOs.
```
mrbean-bremen: This looks good! I just would like to have two minor additions:
- mention that the temp directory that will be created is the one returned by `tempfile.gettempdir()` under each system (as generally there may be several temp directories) 
- add the Windows location of the temp dir (by default `C:\Users\<user>\AppData\Local\Temp`), and maybe add that under MacOS this temp dir is usually something like `/var/folders/<some cryptic ID>/T`

svozza: Ah yes, I knew I was forgetting something: Windows! All good, I'll update accordingly.

svozza: How's that?

PR #479, opened 2019-04-29
Use snapshot of modules to avoid RuntimeError

dunse:
When running code on Python 3.7.3, we get this error intermittently: RuntimeError: dictionary changed size during iteration
Creating a list instead of set appears to solve this issue.

### DIFF
--- a/pyfakefs/fake_filesystem_unittest.py
+++ b/pyfakefs/fake_filesystem_unittest.py
@@ -461,7 +461,7 @@ class Patcher(object):
                 return False
 
         module_names = list(self._fake_module_classes.keys()) + [PATH_MODULE]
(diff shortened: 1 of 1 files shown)


pyfakefs/fake_filesystem_unittest.py
```diff
@@ -461,7 +461,7 @@ def is_fs_function(fct):
                 return False
 
         module_names = list(self._fake_module_classes.keys()) + [PATH_MODULE]
-        for name, module in set(sys.modules.items()):
+        for name, module in list(sys.modules.items()):
```
mrbean-bremen: I don't remeber why I used a set here - there shouldn't be any duplicates, so I guess this will not change the behavior.

PR #491, opened 2019-08-03
Fix scandir DirEntry.path for relative path or fd

nesciens:
The documentation for os.DirEntry.path [1] says that the path attribute
of a DirEntry object is only absolute if scandir()'s path argument was
absolute, and simply the name attribute if scandir()'s path argument was
a filedescriptor. This patch makes pyfakefs do this, and adjusts the
tests.

Relative symlinks are now also covered in the scandir tests.

RealScandirFdTest.test_path_like was failing for me in Python 3.7.3 on
Debian Bullseye.

[1] https://docs.python.org/3/library/os.html#os.DirEntry.path

### DIFF
--- a/pyfakefs/fake_scandir.py
+++ b/pyfakefs/fake_scandir.py
@@ -40,6 +40,7 @@ class DirEntry(BaseClass):
         self._filesystem = filesystem
         self.name = ''
         self.path = ''
+        self._abspath = ''
         self._inode = None
         self._islink = False
         self._isdir = False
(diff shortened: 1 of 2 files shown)


nesciens:
Arg. `pyfakefs.tests.fake_os_test.FakeScandirRelTest` fails on Windows in the AppVeyor build. It seems that the fake scandir gives 0 files when doing scandir with a relative path instead of 2 or 6 depending on the Python version. `pyfakefs.tests.fake_os_test.RealScandirRelTest` (with `Real` instead of `Fake`) succeeds though, so this seems to be something where pyfakefs does not agree with the real behavior on Windows.

Unfortunately I don't have a Windows environment to reproduce this on. I might try under Wine, but might not have enough time to spend on that.

mrbean-bremen:
Thanks for the fix! I'm working under Windows, so I may check and hopefully fix the Windows problem. If you don't mind, I can do this directly on your PR branch, though I'm not sure if I will get to this today.

mrbean-bremen:
Ok, I had a quick look, and the problem seems to be with `relpath`/`abspath` under Windows, where the drive is not correctly taken into account. This is actually a weak point with the Windows implementation in pyfakefs and has to do with the fact that both `C:\` and `\` are handled as root.
I will see what I can do, though this may take a while...

nesciens:
Great, and thanks for the quick review!

PR #494, opened 2019-08-07
Added note about dropping Python 2 support

mrbean-bremen:
@jmcgeheeiv - please check the wording!

### DIFF
--- a/CHANGES.md
+++ b/CHANGES.md
@@ -1,6 +1,18 @@
 # pyfakefs Release Notes
 The release versions are PyPi releases.
 
+## Version 4.0
+
+_Note:_ This version will drop support for Python 2.7. It is planned to be 
+released at the end of 2019 or the beginning of 2020.
+We strongly encourage you to switch to Python 3, if not already done, but 
+if you still need Python 2.7, you can still use the latest pyfakefs 3.x 
+version. 
+
+This version will also remove the old-style API, which has been deprecated in 
+pyfakefs 3.4.
+
+
 ## Version 3.7 (as yet unreleased)
 
 ### Fixes
--- a/CHANGES.md
+++ b/CHANGES.md
@@ -10,9 +10,6 @@ the proposed changes so you can be ready.
   * pyfakefs 4.0 drops support for Python 2.7. If you still need
     Python 2.7, you can still use the latest pyfakefs 3.x version. 
 
-  * pyfakefs 4.0 version will also remove the old-style API, which was
-     deprecated in pyfakefs 3.4.
-
 ## Version 3.7 (as yet unreleased)
 
 ### Fixes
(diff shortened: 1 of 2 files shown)


CHANGES.md
```diff
@@ -1,5 +1,17 @@
...
+## Preview of future Version 4.0
+
+_Note:_ pyfakefs 4.0 is planned for release at the end of 2019 or the beginning
+of 2020. As pyfakefs 4.0 is a major release, we are giving you advance notice of
+the proposed changes so you can be ready.
+
+  * pyfakefs 4.0 drops support for Python 2.7. If you still need
+    Python 2.7, you can still use the latest pyfakefs 3.x version. 
+
+  * pyfakefs 4.0 version will also remove the old-style API, which was
```
jmcgeheeiv: The old-style API needs to be described. This seems like a major breaking change.

mrbean-bremen: Yes, I'm still not completely sure if we shall remove it - I have no idea how frequently it is still used. What do you think?
By "describing" the old API you mean giving a few examples, and/or pointing to the old documentation?

Actually I could check a few packages on GitHub that depend on pyfakefs to get some impression of the API usage. There are 330 repos listed...

mrbean-bremen:
Ok, here are some numbers. I checked GitHub repos that depend on pyfakefs and have at least 10 stars. Some of them don't use pyfakefs, or I couldn't find it. From the remaining 47 repositories:
- 8 use no API at all (e.g no FakeFileSystem methods)
- 23 use the new API
- 16 use the old API, 7 of them have the pyfakefs version fixed to a version <= 3.3, 9 have no fixed version

These last 9 out of 47 repos may have a problem after the change to a version without the old API.
So, given these numbers, what do you think?

jmcgeheeiv:
While I still do not quite understand what you want to remove, could we make something like a facade?

FakeFileSystem = fake_file_system
???

mrbean-bremen:
Ah, sorry! I meant the old-style method names like `CreateFile` and `MakeDirectories` vs the pep-8 ones like `create_file` and `makedirs`. The old ones are deprecated since version 3.4. Though now I think it is probably the best to leave them in...

jmcgeheeiv:
Maybe

FakeFileSystem = fake_file_system
def fake_file_system(...):
    ...

mrbean-bremen:
We already have a facade for the old-style functions, the question was just if we shall remove it after the functions have been deprecated for a few versions. And for me the answer is now probably "no"... 
I also had forgotten that the deprecation warning has been switched on only in pyfakefs 3.5. In version 3.4 there had been no warning for using the old functions.

As for your example - I'm not sure I understand that correctly. `FakeFileSystem` is a class, so its name is just fine...

mrbean-bremen:
Ok, I'm merging this now - we can handle that old API stuff another time.

jmcgeheeiv:
Heh, I wrote that comment in a taxi on the way to a party. Of course class name `FakeFileSystem` is PEP-8 as-is.

mrbean-bremen:
I hope it was a nice party 😛

Issue #496, opened 2019-10-09
Chmod on directories has no effect on readability


ian-h-chamberlain:
**Describe the bug**
After creating a directory with files in it in the fake filesystem, then changing its permissions to disallow `r/w/x`, it is still possible to read the contents of the directory.

In a real filesystem, a directory with perms `000` would raise `PermissionsError` when attempting to read the directory contents.

This is not an instance of #489 – these tests still fail even when run as non-root user.

**How To Reproduce**
```py
"""Minimal reproduction of directory permissions issues"""

import os
import pathlib
import stat

import pytest


@pytest.fixture
def unreadable_dir(fs):
    fs.create_file("/tmp/unreadable/file")
    os.chmod("/tmp/unreadable", 0o000)
    return fs


def test_read_dir_os(unreadable_dir):
    with pytest.raises(OSError):
        os.listdir("/tmp/unreadable")  # Does not raise

    with pytest.raises(OSError):
        os.stat("/tmp/unreadable/file")  # Does not raise


def test_dir_permissions_os(unreadable_dir):
    assert stat.S_IMODE(os.stat("/tmp/unreadable").st_mode) == 0o000


def test_read_dir_pathlib(unreadable_dir):
    path = pathlib.Path("/tmp/unreadable")

    with pytest.raises(OSError):
        path.iterdir()  # Does not raise

    with pytest.raises(OSError):
        (path / "file").stat()  # Does not raise


def test_dir_permissions_pathlib(unreadable_dir):
    path = pathlib.Path("/tmp/unreadable")
    assert stat.S_IMODE(path.stat().st_mode) == 0o000

```

**Your enviroment**
```bash
$ python -c "import platform; print(platform.platform())"
Darwin-18.7.0-x86_64-i386-64bit
$ python -c "import sys; print('Python', sys.version)"
Python 3.6.5 (default, Jan 29 2019, 13:07:11)
[GCC 4.2.1 Compatible Apple LLVM 10.0.0 (clang-1000.11.45.5)]
$ python -c "from pyfakefs.fake_filesystem import __version__; print('pyfakefs', __version__)"
pyfakefs 3.6.1
```

The same symptoms seem to occur when tested on platform
```
Linux-3.10.0-514.el7.x86_64-x86_64-with-centos-7.5.1804-Core
```

mrbean-bremen:
Thanks for the report! This case is indeed not implemented, and has not been noticed so far.
[...]

PR #497, opened 2019-10-15
Added handling of missing directory read access under MacOS

mrbean-bremen:
- changes are for MacOS only, as long as Linux behavior is unclear
- see #496


mrbean-bremen:
@ian-h-chamberlain - I'm going to merge this now. It shall fix the problem under MacOS, while for Linux I still need more information what the behavior depends on.

Issue #501, opened 2019-11-12
disutil commands fails for multiple tests 


bspradling:
**Describe the bug**
While running multiple tests trying to use `disutil` to copy files between directories, every test after the first will fail.

```
Traceback (most recent call last):
  File "tests/plugins/pyfakefs_test.py", line 29, in test_file_copied_again
  File "/usr/local/Cellar/python@2/2.7.15_1/Frameworks/Python.framework/Versions/2.7/lib/python2.7/distutils/dir_util.py", line 163, in copy_tree
  File "/usr/local/Cellar/python@2/2.7.15_1/Frameworks/Python.framework/Versions/2.7/lib/python2.7/distutils/dir_util.py", line 167, in copy_tree
  File "/usr/local/Cellar/python@2/2.7.15_1/Frameworks/Python.framework/Versions/2.7/lib/python2.7/distutils/file_util.py", line 110, in copy_file
DistutilsFileError: can't copy './test2/subdir/1.txt': doesn't exist or not a regular file
```

When running this tests in isolation (one at a time), they both succeed.

**How To Reproduce**
Please provide a unit test or a minimal code snippet that reproduces the 
problem.
```import os
import unittest

from distutils.dir_util import copy_tree, remove_tree
from pyfakefs.fake_filesystem_unittest import TestCase


class DataLossPluginTests(TestCase):

    def setUp(self):
        self.setUpPyfakefs()
        self.bootstrap_fake_file_system(self)

    def test_file_copied(self):
        copy_tree("./test2/", "./test/")
        remove_tree("./test2/")

        self.assertTrue(os.path.isfile('./test/subdir/1.txt'))
        self.assertFalse(os.path.isdir('./test2/'))

    def test_file_copied_again(self):

        self.assertTrue(os.path.isfile('./test2/subdir/1.txt'))

        copy_tree("./test2/", "./test/")
        remove_tree("./test2/")

        self.assertTrue(os.path.isfile('./test/subdir/1.txt'))
        self.assertFalse(os.path.isdir('./test2/'))

    @staticmethod
    def bootstrap_fake_file_system(self):
        self.fs.create_dir("./test/subdir/")
        self.fs.create_dir("./test/subdir2/")
        self.fs.create_dir("./test2/subdir/")

        self.fs.create_file("./test2/subdir/1.txt")


if __name__ == '__main__':
    unittest.main()
```

**Your enviroment**
Please run the following and paste the output.
```bash
[...]

Issue #427, opened 2018-09-07
[3.4.3] Multiple tests in TestCase fail when not run in debug - shutil.make_archive


inkychris:
I'm not sure if this is a bug or an issue with my approach but the following minimal test example only appears to run both tests successfully if run with the debugger using PyCharm 2018.2.3 CE on Windows 10 (Python 3.7.0). 

```py
from pyfakefs.fake_filesystem_unittest import TestCase
import shutil


class Test(TestCase):
    def setUp(self):
        self.setUpPyfakefs()
        self.fs.create_file('src/file_1.txt')
        self.fs.create_file('src/a/file_2.txt')
        shutil.make_archive('archive', 'zip', root_dir='src')

    def test_a(self):
        self.assertTrue(True)

    def test_b(self):
        self.assertTrue(True)
```

Stack trace:
```
.E
======================================================================
ERROR: test_b (scratchpad.Test)
----------------------------------------------------------------------
Traceback (most recent call last):
  File "C:\Dev\venv\lib\site-packages\pyfakefs\fake_filesystem.py", line 1876, in get_object_from_normpath
    target_object = target_object.get_entry(component)
  File "C:\Dev\venv\lib\site-packages\pyfakefs\fake_filesystem.py", line 617, in get_entry
    return self.contents[pathname_name]
KeyError: 'a'

During handling of the above exception, another exception occurred:

Traceback (most recent call last):
  File "C:\Dev\venv\lib\site-packages\pyfakefs\fake_filesystem.py", line 1062, in stat
    entry_path, follow_symlinks, allow_fd=True)
  File "C:\Dev\venv\lib\site-packages\pyfakefs\fake_filesystem.py", line 1921, in resolve
    return self.get_object_from_normpath(self.resolve_path(file_path))
  File "C:\Dev\venv\lib\site-packages\pyfakefs\fake_filesystem.py", line 1878, in get_object_from_normpath
    self.raise_io_error(errno.ENOENT, file_path)
  File "C:\Dev\venv\lib\site-packages\pyfakefs\fake_filesystem.py", line 878, in raise_io_error
    raise IOError(errno, self._error_message(errno), filename)
FileNotFoundError: [Errno 2] No such file or directory in the fake filesystem: '\\a'

During handling of the above exception, another exception occurred:

Traceback (most recent call last):
  File "C:\Dev\scratchpad.py", line 10, in setUp
    shutil.make_archive('archive', 'zip', root_dir='src')
[...]

Issue #500, opened 2019-11-03
Failed tests do not lead to failed builds


mrbean-bremen:
Failing tests do not always lead to failed builds in Travis.CI, for example [this build](https://travis-ci.org/jmcgeheeiv/pyfakefs/jobs/606166603) is shown as succeded while many tests fail.

PR #503, opened 2019-11-15
Unload all modules loaded during the test

mrbean-bremen:
- these module can still hold references to fake modules
- modules already loaded at test start are not affected
- see #501 and #427

### DIFF
--- a/CHANGES.md
+++ b/CHANGES.md
@@ -14,9 +14,9 @@ the proposed changes so you can be ready.
 
 ### Fixes
   * fixed side effect of calling `DirEntry.stat()` under Windows (changed 
-    st_nlink) ([#502](../../issues/502))
-  * fixed a problem related to patching `distutils` functions
-    ([#501](../../issues/501))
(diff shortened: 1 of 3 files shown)


mrbean-bremen:
@jmcgeheeiv - please have a look. This shall fix the mentioned problems, but I would like to have a second opinion on probable problems/side effects this could cause.

jmcgeheeiv:
Well, now I've done it. I changed a comment and now it fails a test. Comments?

mrbean-bremen:
> Well, now I've done it. I changed a comment and now it fails a test. Comments?

You made an error in your correction: `Dereference all force all modules`, so the spell checker found that 😛 
It may also have to do with the PEP8 finding that a line is too long...

I recently (while handling #500) activated the flake8 test (put here by a contributor) so that is fails the build if it has warnings. This is a bit over the top, and I can deactivate it again - it was more of an experiment anyway.

mrbean-bremen:
Thanks!

PR #506, opened 2019-12-19
Add note about accessing real files in pytest

JarnoRFB:
Should give pytest users a way to load test resources that does fit with
the style that pytest is normally used.

### DIFF
--- a/docs/usage.rst
+++ b/docs/usage.rst
@@ -380,6 +380,29 @@ fake filesystem via the argument ``target_path``.
                 # only at this point
                 contents = f.read()
 
+
+When using ``pytest`` another option is to load the contents of the real file
+in a fixture and pass this fixture to the test function **before** passing the ``fs``
+fixture.
+
+.. code:: python
+
+    import pytest
+    import os
+
+    @pytest.fixture
+    def content():
+        fixture_path = os.path.join(os.path.dirname(__file__), 'fixtures')
+        with open(os.path.join(self.fixture_path, 'fixture1.txt') as f:
+            contents = f.read()
+
+        return contents
+
+    def test_using_file_contents(content, fs):
+        fs.create_file("fake/path.txt")
+        assert content != ""
+
+
 Handling mount points
 ~~~~~~~~~~~~~~~~~~~~~
 Under Linux and MacOS, the root path (``/``) is the only mount point created


docs/usage.rst
```diff
@@ -380,6 +380,29 @@ fake filesystem via the argument ``target_path``.
...
+    import pytest
+    import os
+
+    @pytest.fixture
+    def content():
+        fixture_path = os.path.join(os.path.dirname(__file__), 'fixtures')
+        with open(os.path.join(self.fixture_path, 'fixture1.txt') as f:
+            contents = f.read()
+
+        return contents
```
mrbean-bremen: Thanks! Though I think it would be better in this context to use the `fs` methods, as this is about their usage. 
Something like this:
```python
fixture_path = os.path.join(os.path.dirname(__file__), 'fixtures')


@pytest.fixture
def fs_with_content(fs):
    fs.add_real_directory(fixture_path)
    yield fs


def test_using_fixture_path(fs_with_content):
    fixture_file_path = os.path.join(fixture_path, 'fixture1.txt')
    assert os.path.exists(fixture_file_path)
    with open(os.path.join(fixture_path, 'fixture1.txt') as f:
        contents = f.read()  # contents are copied over to the fake fs at this point
        assert contents
```
What do you think?

JarnoRFB: My point was that using `fs.add_real_directory` is not necessary if one acquires the resources before applying the `fs` fixture. This way one does not even have to learn about the API for adding real files to the fake filesystem and can just continue to use the fixtures that rely on real files without changing anything.

mrbean-bremen: True. I actually did exactly this for fixture access in my own code before adding the API. 
Maybe it makes sense to add both approaches to the docs - making a fixture based on the `fs` fixture for setup is not described in the docs so far and may not be obvious to everyone, and using your approach to retrieve file contents before the fs fixture is instantiated is another approach that in some contexts may be easier.

I'm probably going to merge your changes, and add the other stuff later, as you gave me the idea now ;)

mrbean-bremen:
[Here](http://jmcgeheeiv.github.io/pyfakefs/master/usage.html#access-to-files-in-the-real-file-system) are the updated docs.

PR #514, opened 2020-02-08
Added automatic patching of file system functions in default argument

mrbean-bremen:
- works for module-level functions and methods in module-level classes
- this also makes it unnecessary to reload the "tempfile" module


mrbean-bremen:
@jmcgeheeiv - please have a look! I'm not sure about possible side effects...

jmcgeheeiv:
Looks good to me. You're really getting into the corner cases here, but that's what gets stars for the project!

mrbean-bremen:
Thanks - that special handling for `tempfile` has been nagging me for some time...

Issue #531, opened 2020-06-08
Patch known libraries where possible


mrbean-bremen:
There are some libraries that can be made to work with pyfakefs if adding some specific patches. This includes django (file locking, described as an example in the documentation), pandas (at least `from_csv` can be made to work, see #528), and probably more.

It could be helpful to automatically add these patches if the libraries are installed, with the option to disable this if not wanted.

mrbean-bremen:
Has been implemented for Pandas and django, more may be added on demand.

PR #532, opened 2020-06-13
Added partial support for pandas to work with pyfakefs

mrbean-bremen:
- support for read_csv and read_excel (with the default module)
- see #531


mrbean-bremen:
@jmcgeheeiv - please have a look. I will check if can add tests/fixes for more read functions in pandas, but that will not change much.

docs/usage.rst
```diff
@@ -606,8 +618,10 @@ A list of Python modules that are known to not work correctly with
 - the ``Pillow`` image library does not work with pyfakefs at least if writing
   JPEG files (see `this issue <https://github.com/jmcgeheeiv/pyfakefs/issues/529>`__)
 - ``pandas`` (the Python data analysis library) uses its own internal file
-  system access, written in C, and does therefore not work with pyfakefs
-  (see `this issue <https://github.com/jmcgeheeiv/pyfakefs/issues/528>`__)
+  system access, written in C, and does therefore not work with pyfakefs out
```
jmcgeheeiv, line 621: system access written in C. Thus much of ``pandas`` will not work with ``pyfakefs``.
   Having said that, ``pyfakefs`` patches ``pandas`` so that many of the ``read_xxx``
   functions, including ``read_csv`` and ``read_excel``, do work with the fake
   file system. If you use only these functions, ``pyfakefs`` will work with ``pandas``.

mrbean-bremen:
> Note the typesetting for an em dash contains no spaces

Thanks--learned something new! For me the missing spaces look a bit awkward, but that may be a matter of what I am used to (in German, there are spaces around the em dash, as far as I know). I usually don't think about punktuation at all.  
I also found this in the [punktuation guide](https://www.thepunctuationguide.com/em-dash.html):
> The em dash is typically used without spaces on either side, and that is the style used in this guide. Most newspapers, however, set the em dash off with a single space on each side.

I will add you changesr and merge. I will wait a bit (in case new bugs come in, or I think of something else to do...), and probably make another another release (4.1), if that is ok with you.

jmcgeheeiv:
Now tell that punk daughter of yours that your English punctuation has the
seal of approval from a native speaker.

Yes, when I started using the em dash without spaces it did look strange to
me. Check the German usage--you may be surprised!

PR #538, opened 2020-07-12
Skip test_write_excel if openpyxl is not installed

mgorny:
test_write_excel fails if pandas are installed but openpyxl is not.
Adjust the condition around the case appropriately.

### DIFF
--- a/pyfakefs/tests/patched_packages_test.py
+++ b/pyfakefs/tests/patched_packages_test.py
@@ -28,6 +28,11 @@ try:
 except ImportError:
     xlrd = None
 
+try:
+    import openpyxl
+except ImportError:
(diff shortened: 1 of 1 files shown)


mrbean-bremen:
@mgorny - thanks for the patch - can you say on what system you had the problem? In all tested systems (in CI tests) `xlrd` is used as default, so `openpyxl` was not used. It would probably make sense to test both engines explicitely instead..

mgorny:
On Gentoo, with pandas 1.0.4 and xlrd 1.2.0 installed. I think xlrd supports only reading (there's xlwt for writing but I think it's obsolete). It's possible that openpyxl started being used for writing in newer pandas.

PR #587, opened 2021-02-11
Remove @DeprecationWarning as it isn't a decorator

gpshead:
`tearDownPyfakefs` is a no-op legacy API, it was given @DeprecationWarning as a decorator... which broke it rather than making it a no-op callable as that isn't a decorator.

This fixes:

```
    self.tearDownPyfakefs()
TypeError: 'DeprecationWarning' object is not callable
```

in code updating to a modern pyfakefs that still calls it.


mrbean-bremen (approved):
Thanks! This has been there since 2017 (I obviously made a mistake here), and has not been noticed before.

PR #594, opened 2021-05-10
fix breaking pathlib change in 3.10 beta1

carlmontanari:
Hello!

Sorry for the PR outa the blue w/out an issue to discuss -- had some time and figured I'd take a crack at it. I added 3.10 beta1 to my ci for some projects using pyfakefs and my builds are failing. Did a bit of sleuthing and it seems there were a few changes to Pathlib for this release. The main issue/thread seems to be [here](https://github.com/python/cpython/pull/19342) -- and associated [commit](https://github.com/python/cpython/commit/2219187cab6bca009c42b63b2f4c30b5b10c916d). This [commit](https://github.com/python/cpython/commit/abf964942f97f6489360a75fd57b5e4f41c75f57) also appears to be related.

I *think* (hope?!) that this fix is backwards compatible. Tests in the docker container all seem happy, and I've tested this locally on my Mac w/ 3.10 beta1 and the tests that are/were failing in CI pass like normal. For reference, [here](https://github.com/scrapli/scrapli_cfg/runs/2547508032?check_suite_focus=true) is my failed tests (the `build_posix (ubuntu-latest, 3.10.0-beta.1)` job). 

The fix itself is basically to ignore that `init` argument (the pr/issue on Cpython seems to say this is mostly legacy and (obviously) going away anyway). Then allowing the wrapped stat method (well I guess everything... which is perhaps not ideal?) to accept kwargs as well. Since `_init` method is no longer called, I removed that and just assigned that `_accessor` to the `_fake_accessor` directly after the `_from_parts` calls. 

Hope that all makes sense -- have been happily using pyfakefs (thanks for the project btw!) a while but the guts of it are probably a bit over my head :)

Carl

### DIFF
--- a/.github/workflows/pythonpackage.yml
+++ b/.github/workflows/pythonpackage.yml
@@ -37,7 +37,7 @@ jobs:
     steps:
     - uses: actions/checkout@v2
     - name: Set up Python ${{ matrix.python-version }}
-      uses: actions/setup-python@v1
+      uses: actions/setup-python@v2
       with:
         python-version: ${{ matrix.python-version }}
 
--- a/.github/workflows/pythonpackage.yml
+++ b/.github/workflows/pythonpackage.yml
@@ -29,7 +29,7 @@ jobs:
       fail-fast: false
       matrix:
         os: [ubuntu-latest, macOS-latest, windows-2016]
-        python-version: [3.6, 3.7, 3.8, 3.9]
+        python-version: [3.6, 3.7, 3.8, 3.9, 3.10.0-beta.1]
         include:
           - python-version: pypy3
             os: ubuntu-latest
--- a/.github/workflows/pythonpackage.yml
+++ b/.github/workflows/pythonpackage.yml
@@ -79,18 +79,28 @@ jobs:
       shell: bash
     - name: Install extra dependencies
       run: |
-        pip install -r extra_requirements.txt
+        # some extra dependencies are not avaialble in 3.10 Beta yet
+        # so we exclude it from all tests on extra dependencies
+        if [[ '${{ matrix.python-version }}' != '3.10.0-beta.1' ]]; then
+          pip install -r extra_requirements.txt
+        fi
+      shell: bash
     - name: Run unit tests with extra packages as non-root user
       run: |
-        python -m pyfakefs.tests.all_tests
+        if [[ '${{ matrix.python-version }}' != '3.10.0-beta.1' ]]; then
+          python -m pyfakefs.tests.all_tests
+        fi
+      shell: bash
     - name: Run pytest tests
       run: |
-        export PY_VERSION=${{ matrix.python-version }}
-        $GITHUB_WORKSPACE/.github/workflows/run_pytest.sh
+        if [[ '${{ matrix.python-version }}' != '3.10.0-beta.1' ]]; then
+          export PY_VERSION=${{ matrix.python-version }}
+          $GITHUB_WORKSPACE/.github/workflows/run_pytest.sh
+        fi
       shell: bash
     - name: Run performance tests
       run: |
-        if [[ '${{ matrix.os  }}' != 'macOS-latest' ]]; then
+        if [[ '${{ matrix.os  }}' != 'macOS-latest' && '${{ matrix.python-version }}' != '3.10.0-beta.1' ]]; then
           export TEST_PERFORMANCE=1
           python -m pyfakefs.tests.performance_test
         fi
--- a/pyfakefs/fake_filesystem.py
+++ b/pyfakefs/fake_filesystem.py
@@ -114,7 +114,7 @@ from pyfakefs.fake_scandir import scandir, walk
 from pyfakefs.helpers import (
     FakeStatResult, BinaryBufferIO, TextBufferIO,
     is_int_type, is_byte_string, is_unicode_string,
-    make_string_path, IS_WIN, to_string, matching_string
+    make_string_path, IS_WIN, to_string, matching_string, real_encoding
 )
 from pyfakefs import __version__  # noqa: F401 for upwards compatibility
 
@@ -293,7 +293,7 @@ class FakeFile:
         if st_mode >> 12 == 0:
             st_mode |= S_IFREG
         self.stat_result.st_mode = st_mode
-        self.encoding = encoding
+        self.encoding = real_encoding(encoding)
         self.errors = errors or 'strict'
         self._byte_contents = self._encode_contents(contents)
         self.stat_result.st_size = (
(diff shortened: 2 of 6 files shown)


pyfakefs/fake_pathlib.py
```diff
@@ -722,7 +718,8 @@ def __new__(cls, *args, **kwargs):
         if cls is RealPathlibModule.Path:
             cls = (RealPathlibModule.WindowsPath if os.name == 'nt'
                    else RealPathlibModule.PosixPath)
-        self = cls._from_parts(args, init=True)
+        self = cls._from_parts(args)
+        self._accessor = _fake_accessor
         return self
```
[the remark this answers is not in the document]
carlmontanari: Fixed!

.github/workflows/pythonpackage.yml
```diff
@@ -28,7 +29,7 @@ jobs:
       fail-fast: false
       matrix:
         os: [ubuntu-latest, macOS-latest, windows-2016]
-        python-version: [3.6, 3.7, 3.8, 3.9]
+        python-version: [3.6, 3.7, 3.8, 3.9, 3.10.0-beta.1]
```
mrbean-bremen, line 32: Thanks! I think you need `3.10-dev` here - GH actions won't understand that.

mrbean-bremen: > looks like we may need to update the python actions to v2 as well

Ah, didn't realize that, thanks! Though currently some tests are failing for all versions, something is not correct yet...

mrbean-bremen: >  I'll try to take a peak at the failures in later today/tomorrow

Thanks - I can also have another look if you want, but probably not before tomorrow evening (noon for you).

carlmontanari:
Just pushed a commit that restores the `_init` method.... for whatever reason it seems that its working. I would have assumed because it was being called in `__new__` we would have been hitting these lines where we set the accessor and closed anyway?   Can circle back on this tomorrow with fresh eyes if that didn't work for some reason :D

pyfakefs/fake_filesystem.py
```diff
@@ -1177,9 +1177,12 @@ def stat(self, entry_path, follow_symlinks=True):
...
         # stat should return the tuple representing return value of os.stat
-        file_object = self.resolve(
-            entry_path, follow_symlinks,
-            allow_fd=True, check_read_perm=False)
+        try:
+            file_object = self.resolve(
+                entry_path, follow_symlinks,
+                allow_fd=True, check_read_perm=False)
+        except TypeError:
+            file_object = self.resolve(entry_path)
```
carlmontanari, line 1185: this looks like another update that was made to pathlib (https://github.com/python/cpython/blob/main/Lib/pathlib.py#L1052)

not sure this is a particularly good way to handle it but it *does* seem to work

pyfakefs/fake_pathlib.py
```diff
@@ -131,9 +132,17 @@ def lchmod(self, pathobj, mode):
...
+        # why? and shouldnt this not be *os* -- but something patched (FakeOsModule probably)?
+        readlink = staticmethod(os.readlink)
+        pass
 
     utime = _wrap_strfunc(FakeFilesystem.utime)
 
+    # same comment as above for these two
+    realpath = staticmethod(os.path.realpath)
+    getcwd = staticmethod(os.getcwd)
+
```
carlmontanari, line 145: this and the above section I am 99% sure I am doing something wrong/bad here -- but the gist of this is it looks like we needed to add some methods to the FakeAccessor object -- I am *assuming*(?) its not a great idea to just attach `os` methods here because we would want to have the accessor using the patched methods? But this does seem to get more of the tests passing so figured id commit it to try to be as helpful as I can... (even if it isn't much :))

carlmontanari:
Last push for the day because I am def out of my element here :) I left some comments in the files changed section about things im not super sure about and/or I think are probably wrong. This does get tox tests passing locally but some of it feels fairly wrong. Hopefully this can at least save ya some time on figuring stuff out!

carlmontanari:
@mrbean-bremen Sorry I didn't get a chance to look today either, and sorry for my barrage of not working commits :). If you dont get to it first I will try to spend some more time looking at this this weekend but I think I am in over my head! Whatever is easiest for you w/ respect to the PR I'm totally good w/! I'm glad to have at least helped a little bit!

mrbean-bremen:
@carlmontanari -- thanks! There is no need for you to spend more time on this, I can take it from here.  As I wrote, your contribution has already been helpful enough. I'll work on your branch then if you don't mind and will see if I can finish it soon -- maybe at the weekend.

pyfakefs/fake_filesystem.py
mrbean-bremen: Will leave it as is, I think it is good enough for now.

pyfakefs/fake_pathlib.py
mrbean-bremen: I changed it accordingly. Regarding `getcwd` - if using `os` inside the function, as I do it now, it will be patched, so it will behave correctly.

mrbean-bremen:
@carlmontanari - I got it to work now, I'll let you have a look and check if something is missing from your POV before I'm going to merge it. Didn't have to wait until the weekend, as today was a holiday here...

pyfakefs/fake_pathlib.py
```diff
@@ -53,8 +53,9 @@ def init_module(filesystem):
 
 def _wrap_strfunc(strfunc):
     @functools.wraps(strfunc)
-    def _wrapped(pathobj, *args):
-        return strfunc(pathobj.filesystem, str(pathobj), *args)
+    def _wrapped(pathobj, *args, **kwargs):
+        p = pathobj.filesystem if hasattr(pathobj, "filesystem") else pathobj
```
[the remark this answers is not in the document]
mrbean-bremen: This change was actually not needed - it happened because the fake accessor was not set (why I had to implement the two class methods to call `_init`).

pyfakefs/fake_pathlib.py
mrbean-bremen: I removed `realpath` as it doesn't seem to be really needed, but maybe I'm wrong here.

carlmontanari:
> @carlmontanari - I got it to work now, I'll let you have a look and check if something is missing from your POV before I'm going to merge it. Didn't have to wait until the weekend, as today was a holiday here...

You rock! A quick check and looks like all my tests are passing on 3.10 locally. Will try to get a branch updated to remove my skips to let it run on 3.10 on ubuntu latest too (locally am macOS).

mrbean-bremen (approved):

Issue #609, opened 2021-06-07
pylint complains about `fs` fixture name


AustinScola:
I'm using the pytest `pyfakefs` fixture in a project. The project is linted with `pylint`. `pylint` warns `C0103: Argument name "fs" doesn't conform to snake_case naming style (invalid-name)`. I know that I could add `fs` to the list of `good-names` for `pylint`, but I would prefer not to do that because then the name would be accepted everywhere. I could also add `# pylint: disable=invalid-name` everywhere I use the fixture, but that isn't very clean.

Could we add another name for the `pytest` fixture? Something like `fake_filesystem`?

mrbean-bremen:
Hm, `fs` is indeed a bit short, but it's also handy and it's quite clear that it stands for `file system`. I want to avoid polluting the fixture namespace with aliases that cannot be disabled, so I think the way to go here is to define your own fixture:

**conftest.py**
```py
@pytest.fixture
def fake_filesystem(fs):  # pylint:disable=invalid-name
    yield fs
```
and use `fake_filesystem` everywhere else. We _could_ add this as a default, but at the moment I'm more inclined not to do this. 

@jmcgeheeiv - what do you think?

jmcgeheeiv:
I like @mrbean-bremen's workaround. We could add it to our pytest template so we avoid polluting the namespace, yet everybody sees it and can use it to avoid the PyLint issue if they like.

AustinScola:
I like @mrbean-bremen's workaround too. Not polluting the namespace sounds like a good reason. Thanks for the suggestion!

As a side note, something else I thought of is that there might be some weird behavior if both the `fs` and `fake_filesystem` fixtures are used in the same unit test?

mrbean-bremen:
> there might be some weird behavior if both the fs and fake_filesystem fixtures are used in the same unit test

No, they will both point to the same fake fs and can be used interchangeably:
```py
def test_both_fixtures(fs, fake_filesystem):
    fake_filesystem.create_file('/var/data/xx1.txt')
    fs.create_file('/var/data/xx2.txt')
    assert os.path.exists('/var/data/xx1.txt')
    assert os.path.exists('/var/data/xx2.txt')
    assert fs == fake_filesystem
```

mrbean-bremen:
I'm converting this into a discussion Q/A issue.

PR #610, opened 2021-06-07
Add example for alias for fs fixture

mrbean-bremen:
- see #609

As proposed by @jmcgeheeiv.

### DIFF
--- a/docs/usage.rst
+++ b/docs/usage.rst
@@ -50,13 +50,19 @@ plugins, so you can just use it:
        fs.create_file('/var/data/xx1.txt')
        assert os.path.exists('/var/data/xx1.txt')
 
-If you don't like the ``fs`` fixture name (``pylint`` may complain about the
-name), you may define your own alias in your ``conftest.py``:
+If you are bothered by the ``pylint`` warning,
+``C0103: Argument name "fs" doesn't conform to snake_case naming style
+(invalid-name)``,
+you can define a longer name in your ``conftest.py`` and use that in your
+tests:
 
 .. code:: python
 
     @pytest.fixture
-    def fake_filesystem(fs):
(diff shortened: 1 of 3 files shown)


docs/usage.rst
```diff
@@ -50,6 +50,16 @@ plugins, so you can just use it:
        fs.create_file('/var/data/xx1.txt')
        assert os.path.exists('/var/data/xx1.txt')
 
+If you don't like the ``fs`` fixture name (``pylint`` may complain about the
```
jmcgeheeiv, line 53: If you are bothered by PyLint warning,
`C0103: Argument name "fs" doesn't conform to snake_case naming style (invalid-name)`,
you can define a longer name in your ``conftest.py`` and use that in your tests:

docs/usage.rst
```diff
@@ -50,6 +50,16 @@ plugins, so you can just use it:
...
+name), you may define your own alias in your ``conftest.py``:
+
+.. code:: python
+
+    @pytest.fixture
+    def fake_filesystem(fs):
```
jmcgeheeiv, line 59: No `# pylint:disable=invalid-name`?

jmcgeheeiv: """Variable name 'fs' causes a PyLint warning. Provide a longer name acceptable
to PyLint for use in tests.
"""

mrbean-bremen: Thanks! Will make the changes tomorrow.

mrbean-bremen: > No # pylint:disable=invalid-name?

We don't use `pylint`, and `flake8` doesn't have problems with this, but we can add it for illustration in this case.

mrbean-bremen: Actually you can ignore this comment - I overlooked that this in the documentation, not in the example code. I have added it as proposed.

PR #615, opened 2021-06-21
Require Python >= 3.6

mdippery:
pyfakefs 4.5 removed support for Python 3.5, but the python_requires metadata was not updated to reflect this, which means that pyfakefs can be installed in a Python 3.5 environment but will not run correctly.


mrbean-bremen (approved):
Thanks, I had missed that!

PR #634, opened 2021-09-08
Change 'is windows_fs' to 'is_windows_fs'

exvertus:
I was reading through the usage doc and noticed it referencing a 'is windows_fs' which looks like it is missing an underscore according to the [code](https://github.com/jmcgeheeiv/pyfakefs/blob/e06472b0671d2f077cc41e9ab9cf38893ef0c97f/pyfakefs/fake_filesystem.py#L929).

### DIFF
--- a/docs/usage.rst
+++ b/docs/usage.rst
@@ -780,10 +780,10 @@ before adding any files.
 Setting the ``os`` attributes changes a number of ``pyfakefs.FakeFilesystem``
 attributes, which can also be set separately if needed:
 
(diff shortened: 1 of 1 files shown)


mrbean-bremen (approved):
Thanks! Good to know that somebody actually reads the documentation...

exvertus:
> Thanks! Good to know that somebody actually reads the documentation...

You're welcome, thanks for accepting. Have a good day!

Issue #664, opened 2022-02-03
binary mode doesn't take an encoding argument


yaythomas:
The actual `open()` function does not allow you to pass an encoding when in binary mode:

```
>>> open('filename', 'wb', encoding='utf-8')
ValueError: binary mode doesn't take an encoding argument
>>> open('arb', 'b', encoding='utf-8')
ValueError: binary mode doesn't take an encoding argument
```

Assuming this is not an intentional divergence from `open()` behaviour, I think the fix would be to insert this:

```python
if binary and encoding:
    raise ValueError("binary mode doesn't take an encoding argument")
```

just after here:
https://github.com/jmcgeheeiv/pyfakefs/blob/4dc8615957dadfc64d6a8f864ee93c7c65da38b5/pyfakefs/fake_filesystem.py#L5574

Before I prepare a PR, I thought I'd ask whether you think this is suitable first?

P.S this is an AMAZING project - it's my favourite new thing in my tool-belt! Thank you so much for all the work and care that went into this complex wonderful beast!

mrbean-bremen:
Thanks :) And good catch - no this wasn't meant to diverge, we just overlooked it.
You mean if its a suitable first PR? Sure, go ahead!

yaythomas:
By suitable I meant more whether my suggested simplistic fix was comically naive and unaware of other complexities :-)

PR incoming. . .

mrbean-bremen:
The test coverage is pretty ok here, so it is not very likely that you will break something if the tests pass. And I guess you will add a test of your own for the change...

mrbean-bremen:
Thank you, nice work!

yaythomas:
Pleasure was mine! Thanks for a smooth contribution experience!

Issue #542, opened 2020-08-06
DeprecationWarning: DocumentLS.async is deprecated; use DocumentLS.async_


jbwdevries:
Also, 'DeprecationWarning: Document.async is deprecated; use Document.async_'

```
venv/lib/python3.6/site-packages/pyfakefs/pytest_plugin.py:40: 
_ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _
venv/lib/python3.6/site-packages/pyfakefs/fake_filesystem_unittest.py:591: in setUp
    self._find_modules()
venv/lib/python3.6/site-packages/pyfakefs/fake_filesystem_unittest.py:559: in _find_modules
    for f, i, d in self._def_values(fct):
venv/lib/python3.6/site-packages/pyfakefs/fake_filesystem_unittest.py:507: in _def_values
    predicate=inspect.isfunction):
/usr/lib/python3.6/inspect.py:342: in getmembers
    value = getattr(object, key)
/usr/lib/python3.6/xml/dom/xmlbuilder.py:346: in __get__
    self.warn(cls)
```

The item being inspected is `xml.dom.xmlbuilder.DocumentLS`

test_foo.py:
```
def test_my_func(fs):
    assert True
```

requirements.txt:
```
pyfakefs==4.1.0
pytest==6.0.1
pytest-cov==2.9.0
```

Makefile:
```
setup:
	python3 -m venv venv
	venv/bin/python -m pip install -r requirements.txt

test:
	venv/bin/python -m pytest
```

python 3.6.9

mrbean-bremen:
Thanks for the report.

Hm, if some module issues a warning while inspecting one of its attributes, the only thing we can do here is just suppress all warnings, or at least deprecation warnings at this point, as we have no control over which modules are loaded.  
This should be ok... I will probably check this over the weekend.

[...]

PR #665, opened 2022-02-03
Open binary no encoding

yaythomas:
On open() binary mode should not allow an `encoding` argument. See #664.

If encoding is passed with binary mode, raise: `ValueError(binary mode doesn't take an encoding argument)`

Add unit test helper to check the open() constructor call. Test for both binary and text mode with encoding, where binary fails and text passes.

### DIFF
--- a/CHANGES.md
+++ b/CHANGES.md
@@ -7,7 +7,9 @@ The released versions correspond to PyPi releases.
 * correctly handle file system space for files opened in write mode
   (see [#660](../../issues/660))
 * correctly handle reading/writing pipes via file
-  (see [#661](../../issues/661)) 
+  (see [#661](../../issues/661))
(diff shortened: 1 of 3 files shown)


yaythomas:
When I did my local due diligence to pre-check that the CI checks will all pass, I got a mypy error:

```
pyfakefs/fake_filesystem_unittest.py:750: error: Module has no attribute "_HAS_FCOPYFILE"  [attr-defined]
Found 1 error in 1 file (checked 23 source files)
```

I'm pretty sure this is not related to the changeset under consideration.

It _might_ have something to do with issue #542.

mrbean-bremen (approved):
Thanks, this looks good! Can you add an entry to the release notes (CHANGES.md)?

PR #671, opened 2022-03-28
PyTest -> pytest

okken:
`README.md` and `intro.rst` both had a reference to pytest as "PyTest".  
pytest.org only refers to pytest in all lowercase, and the core team prefers that spelling. 

Also, thanks for including a pytest fixture in pyfakefs. Super cool.


mrbean-bremen (approved):
Thanks! This has been there for a while...

PR #729, opened 2022-10-08
Add automatic release on PyPI

mrbean-bremen:
- triggered on publishing a GH release
- includes creation of release documentation
- add PR templates, adapt issue templates
- minor documentation changes

### DIFF
--- a/.github/workflows/release-deploy.yml
+++ b/.github/workflows/release-deploy.yml
@@ -26,16 +26,16 @@ jobs:
     - name: Build package
       run: |
         python -m pip install --upgrade pip
-        python -m pip install wheel
+        python -m pip install build
         cd main
         echo "RELEASE_VERSION=$(cat pyfakefs/_version.py | cut -d "'" -f 2)" >> $GITHUB_ENV
-        python setup.py sdist bdist_wheel
+        python -m build
 
     - name: Publish package to PyPI
       uses: pypa/gh-action-pypi-publish@master
       with:
         user: __token__
-        password: ${{ secrets.PYPI_API_TOKEN }}
+        password: ${{ secrets.PYPI_TOKEN }}
 
     - name: Create release docs
       run: |
--- a/.github/workflows/release-deploy.yml
+++ b/.github/workflows/release-deploy.yml
@@ -1,0 +1,65 @@
+name: release-deploy
+
+on:
+  release:
+    types: [ published ]
+
+jobs:
+
+  deploy:
+    runs-on: ubuntu-latest
+    strategy:
+      fail-fast: true
+      matrix:
+        python-version: [ '3.10' ]
+
+    steps:
+    - uses: actions/checkout@v2
+      with:
+        path: main
+
+    - name: Set up Python ${{ matrix.python-version }}
+      uses: actions/setup-python@v2
+      with:
+        python-version: ${{ matrix.python-version }}
+
+    - name: Build package
+      run: |
+        python -m pip install --upgrade pip
+        python -m pip install wheel
+        cd main
+        echo "RELEASE_VERSION=$(cat pyfakefs/_version.py | cut -d "'" -f 2)" >> $GITHUB_ENV
+        python setup.py sdist bdist_wheel
+
+    - name: Publish package to PyPI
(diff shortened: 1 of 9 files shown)


.github/workflows/release-deploy.yml
```diff
@@ -0,0 +1,65 @@
...
+      with:
+        python-version: ${{ matrix.python-version }}
+
+    - name: Build package
+      run: |
+        python -m pip install --upgrade pip
+        python -m pip install wheel
+        cd main
+        echo "RELEASE_VERSION=$(cat pyfakefs/_version.py | cut -d "'" -f 2)" >> $GITHUB_ENV
+        python setup.py sdist bdist_wheel
```
nicoddemus, line 32: Modern practices suggest using `build`:

https://github.com/pytest-dev/pytest-mock/blob/b6ea52f32fd0221b0172dd46f3c8d9e04b139d35/.github/workflows/deploy.yml#L19-L25

.github/workflows/release-deploy.yml
```diff
@@ -0,0 +1,65 @@
...
+
+    - name: Publish package to PyPI
+      uses: pypa/gh-action-pypi-publish@master
+      with:
+        user: __token__
+        password: ${{ secrets.PYPI_API_TOKEN }}
```
nicoddemus, line 38: If it is all the same to you, I suggest using `PYPI_TOKEN` for consistency with other repositories in pytest-dev (but this is not 100% consistent either, so feel free to disregard this).

.github/workflows/release-deploy.yml
```diff
@@ -0,0 +1,65 @@
...
+
+    - name: Checkout gh-pages
+      uses: actions/checkout@v2
+      with:
+        ref: gh-pages
+        path: doc
+
+    - name: Copy and commit release documentation
+      run: |
+        cp -r main/gh-pages/* doc/release
```
nicoddemus, line 56: Ahh interesting, the docs are published in gh-pages.

In the future I suggest to use readthedocs.io, which builds the docs automatically for releases and even PRs for preview. 👍

nicoddemus (approved):

.github/workflows/release-deploy.yml
mrbean-bremen: Sure, will change this.

.github/workflows/release-deploy.yml
nicoddemus, line 32:
```suggestion
        python -m pip install build
        cd main
        echo "RELEASE_VERSION=$(cat pyfakefs/_version.py | cut -d "'" -f 2)" >> $GITHUB_ENV
        python -m build
```

mrbean-bremen: Thanks, that's better - will change this.

.github/workflows/release-deploy.yml
mrbean-bremen: Until now, I have always done it this way (including in `pytest-order`), but I will have a look at readthedocs in this case, thanks. Not for this release, though...

.github/workflows/release-deploy.yml
mrbean-bremen: Already had done this locally 😄

.github/workflows/release-deploy.yml
mrbean-bremen: The preview feature is indeed something I am missing in gh-pages, so I will definitely check this out!

mrbean-bremen: Actually, setting up ReadTheDocs is very easy, as it turned out - I will probably change this before the release. Maybe tomorrow if I get to it... The PR preview is too good a feature to miss, and as far as I understand, there is literally nothing to do in the release workflow (apart from removing the GH Pages stuff again). Thanks for the tip!

nicoddemus: Sure thing!

The preview is not enabled by default, you need to go into advanced settings and enable it:

image

mrbean-bremen: Thanks, I noticed this. A bit odd that this cannot be configured in the configuration file...

Issue #752, opened 2022-11-26
How to use with setUpClass()


buhtz:
From the documentation it is not clear how I can use PyFakeFS in a setUpClass().

`setuppyfakefs()` only works with a `self` instance from a `setUp()` method.

My current solution is to use a `Patcher` object in `setUpClass()` like this:

```
def My(unittest.TestCase):
    @classmethod
    def tearDownClass(cls):
        cls._patcher.tearDown()

    @classmethod
    def setUpClass(cls):
        cls._patcher = pyfakefs_ut.Patcher()
        cls._patcher.setUp()
```

Maybe you know a better way?

You could improve the "Usage" section of the docu with an situation using `setUpClass()`.

My use-case:
I do create some testfiles in the fake filesystem. Because I only want to read them in the tests I need to create them only once for all tests. Creating them for each test again and again is ressourc inefficient.

mrbean-bremen:
No, there is no specific support for that. It is assumed that the setup is fast enough to happen in each test, and so far, this has not been an issue.
Your implementation is fine as is, and I'm not sure if we need more than that, given that this is the first time this use case ever came up. Supporting this specifically would mean just a thin wrapper around that implementation, and I'm not sure if it's worth it.
You are right about the documentation - we could that use case to the usage section.

mrbean-bremen:
Actually, on a second thought, it makes probably sense to add a class method for setup. As I recently changed the implementation to allow only one patched instance (this was made for session or module scoped fixtures in pytest), this will allow to access the fake filesystem via `self.fs` in the tests as usual.

mrbean-bremen:
I had another look at this, and while I can add a convenience function for class setup (I did this locally), it may not behave as expected. It would ehave the same way as creating the patcher in the class setup, as shown in the original comment. That would also mean that any changes to the file system in a single test would persist, so the test would influence each other.
Ideally, we would need to capture the state at test begin and reset the file system to that after test end, but I'm not sure how feasible this would be.
@buhtz, what do you think?

buhtz:
> That would also mean that any changes to the file system in a single test would persist, so the test would influence each other. Ideally, we would need to capture the state at test begin and reset the file system to that after test end, but I'm not sure how feasible this would be. @buhtz, what do you think?

[...]

PR #758, opened 2022-12-07
Add setUpClassPyfakefs convenience method

mrbean-bremen:
- only available from Python 3.8 onwards
- also add class-scoped 'fs_class' fixture
- handle patcher for doc tests separately
- see #752 

#### Tasks
- [x] Fix or feature added
- [ ] Unit tests passing


docs/usage.rst
```diff
@@ -74,6 +76,39 @@ with the fake file system functions and modules:
...
+If your setup is the same for all tests in a class, you can use the class setup
+method ``setUpClassPyfakefs`` instead:
+
+.. code:: python
+
+    from pyfakefs.fake_filesystem_unittest import TestCase
+
+
+    class ExampleTestCase(TestCase):
+        def setUpClass(cls):
```
buhtz, line 88: Shouldn't there be an `@classmethod` as decorator?

docs/usage.rst
```diff
@@ -74,6 +76,39 @@ with the fake file system functions and modules:
...
+            self.setUpClassPyfakefs()
```
buhtz, line 89: IMHO not `self.` but `cls.`

docs/usage.rst
```diff
@@ -74,6 +76,39 @@ with the fake file system functions and modules:
...
+            cls.file_path = "/test/file.txt"
```
buhtz, line 90: Why so complicated? ;)
Just do

`pathlib.Path('/test/file.txt').touch()`

Isn't one of the great things of your package that I can do all file operations as usual because your package does the patching in background for me. No need to use `fake_fs()` or `create_file()`. Or do I misunderstand something?

buhtz (commented):
I just gave comments about improving the example code in the docstring.

buhtz:
The GitHub review mechanic do not show the correct lines here in my comments. But I hope you can understand them also.

docs/usage.rst
mrbean-bremen: Yes, looks like I botched the example... thanks, will correct this!

docs/usage.rst
mrbean-bremen: True of course.

docs/usage.rst
mrbean-bremen: No, you understand correctly, this is just to demonstrate that you _can_ access the fake fs from there, say to setup things that cannot be done in the real fs (like setting the file system size). But I may change this to make this more clear.

mrbean-bremen: I should have waited for a review before merging this - I tend to overlook simple stuff like that myself, especially in the evenings after my day job... next time.

PR #762, opened 2022-12-21
Update GH actions to Python 3.11 final

browniebroke:
#### Describe the changes
Python 3.11 final is released, so update the test suite to use it instead of the `-dev` version.

#### Tasks
- [ ] Fix or feature added
- [ ] Entry to release notes added
- [ ] Unit tests passing

### DIFF
--- a/.github/workflows/testsuite.yml
+++ b/.github/workflows/testsuite.yml
@@ -10,7 +10,7 @@ jobs:
       fail-fast: false
       matrix:
         os: [ubuntu-latest, macOS-latest, windows-latest]
-        python-version: [3.7, 3.8, 3.9, "3.10", "3.11-dev"]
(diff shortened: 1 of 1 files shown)


buhtz:
I would suggest to completely remove 3.7 because it is in security phase only and will be droped next year by Python.
Also 3.8 and 3.9 are in security phase only. But they are supported fix security fixes (only) for some years more.

mrbean-bremen (approved):
Thanks! Could you also add `py311` to `tox.ini`, while at it?

mrbean-bremen:
> I would suggest to completely remove 3.7 because it is in security phase only

It is still supported, so I see no reason to remove the support now. Also, there are no 3.8-only features that we are waiting for to use in `pyfakefs`, so there is no rush...

mrbean-bremen (approved):
Thanks!

PR #764, opened 2022-12-21
Add Python 3.12 to the test suite

browniebroke:
#### Describe the changes
I know this is quite early 😄 but I was trying to test one of my packages to see if it worked and noticed that pyfakefs was returning some errors. I fixed a few issues to make it work on my machine (macOS) but it looks like it now fails due to pandas not installing yet. Also Windows fails but I don't have such a machine, so I might not be the best person to fix these...

Let me know how I should proceed with this. I understand if you want to close or leave this pull request open until all the jobs pass. I'm also happy to split the fixes and leave the CI updates out for now.

However, I do think that having basic support in this package might help others packages to add Python 3.12 compatibility.

#### Tasks
- [x] Fix or feature added
- [x] Unit tests passing

### DIFF
--- a/.github/workflows/testsuite.yml
+++ b/.github/workflows/testsuite.yml
@@ -10,7 +10,7 @@ jobs:
       fail-fast: false
       matrix:
         os: [ubuntu-latest, macOS-latest, windows-latest]
-        python-version: [3.7, 3.8, 3.9, "3.10", "3.11"]
+        python-version: [3.7, 3.8, 3.9, "3.10", "3.11", "3.12-dev"]
         include:
           - python-version: "pypy-3.7"
             os: ubuntu-latest
--- a/.github/workflows/testsuite.yml
+++ b/.github/workflows/testsuite.yml
@@ -6,14 +6,26 @@ on:
 jobs:
   tests:
     runs-on: ${{ matrix.os }}
+    continue-on-error: ${{ matrix.experimental }}
     strategy:
       fail-fast: false
       matrix:
         os: [ubuntu-latest, macOS-latest, windows-latest]
-        python-version: [3.7, 3.8, 3.9, "3.10", "3.11", "3.12-dev"]
+        python-version: [3.7, 3.8, 3.9, "3.10", "3.11"]
+        experimental: [false]
         include:
-          - python-version: "pypy-3.7"
-            os: ubuntu-latest
+          - os: ubuntu-latest
+            python-version: "pypy-3.7"
+            experimental: false
+          - os: ubuntu-latest
+            python-version: "3.12-dev"
+            experimental: true
+          - os: macOS-latest
+            python-version: "3.12-dev"
+            experimental: true
+          - os: windows-latest
+            python-version: "3.12-dev"
+            experimental: true
 
     steps:
     - uses: actions/checkout@v2
--- a/.github/workflows/testsuite.yml
+++ b/.github/workflows/testsuite.yml
@@ -68,7 +68,6 @@ jobs:
         python -m pyfakefs.tests.all_tests
       shell: bash
     - name: Run performance tests
-      if: ${{ python-version == "3.12-dev" }}
       run: |
         if [[ '${{ matrix.os  }}' != 'macOS-latest' ]]; then
           export TEST_PERFORMANCE=1
--- a/.github/workflows/testsuite.yml
+++ b/.github/workflows/testsuite.yml
@@ -58,12 +58,12 @@ jobs:
         fi
       shell: bash
     - name: Install extra dependencies
-      if: ${{ python-version == "3.12-dev" }}
+      if: ${{ python-version != "3.12-dev" }}
       run: |
         pip install -r extra_requirements.txt
       shell: bash
     - name: Run unit tests with extra packages as non-root user
-      if: ${{ python-version == "3.12-dev" }}
+      if: ${{ python-version != "3.12-dev" }}
       run: |
         python -m pyfakefs.tests.all_tests
       shell: bash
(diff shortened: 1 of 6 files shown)


pyfakefs/fake_pathlib.py
```diff
@@ -383,6 +385,7 @@ class _FakeWindowsFlavour(_FakeFlavour):
         | {"COM%d" % i for i in range(1, 10)}
         | {"LPT%d" % i for i in range(1, 10)}
     )
+    pathmod = ntpath
```
browniebroke, line 388: This class attribute was present but seemingly unused before https://github.com/python/cpython/pull/95450

pyfakefs/tests/fake_pathlib_test.py
```diff
@@ -632,6 +632,7 @@ def test_symlink_to(self):
         self.assertTrue(path.is_symlink())
 
     @unittest.skipIf(sys.version_info < (3, 8), "link_to new in Python 3.8")
+    @unittest.skipIf(sys.version_info >= (3, 12), "link_to removed in Python 3.12")
```
browniebroke, line 635: https://github.com/python/cpython/blob/0397f040e239c82848f51dedef9af7b9bd99ef55/Lib/pathlib.py#L1222-L1225

browniebroke:
The Windows tests are failing because of this new feature that was added: https://github.com/python/cpython/pull/99548

> `os.DirEntry` now includes an `os.DirEntry.is_junction` method to check if the entry is a junction.

pyfakefs/fake_filesystem.py
```diff
@@ -576,6 +576,11 @@ def path(self) -> AnyStr:
             dir_path = sep.join(names)
         return self.filesystem.absnormpath(dir_path)
 
+    @property
+    def is_junction(self) -> bool:
+        # TODO: implement junctions
+        return False
```
mrbean-bremen, line 582: This should probably return `self.filesystem.is_junction(self.path).

.github/workflows/testsuite.yml
```diff
@@ -6,14 +6,26 @@ on:
...
+            experimental: false
+          - os: ubuntu-latest
+            python-version: "3.12-dev"
+            experimental: true
+          - os: macOS-latest
+            python-version: "3.12-dev"
+            experimental: true
+          - os: windows-latest
+            python-version: "3.12-dev"
+            experimental: true
```
mrbean-bremen, line 28: I would probably just disable the tests with extra requirements for 3.12 (by adding a version check to the respective steps). It is common that larger packages do not support a new Python in early development stages. We can remove the check later.

pyfakefs/fake_filesystem.py
```diff
@@ -3440,6 +3445,22 @@ def islink(self, path: AnyPath) -> bool:
...
+
+        Returns:
+            `False` on posix systems.
+            `True` if path is a junction on Windows.
+
+        Raises:
+            TypeError: if path is None.
+        """
+        # TODO: implement junction on Windows
+        return False
```
mrbean-bremen, line 3462: I think for the time being it is ok to just return `False` here - I would adapt the description accordingly, though. Later, we may add some convenience function to make a junction in `pyfakefs` - as far as I can see, this is not possible in Python yet. We also need to adapt `os.stat`, `os.lstat` and possibly other functions that will support junctions.

pyfakefs/tests/fake_pathlib_test.py
mrbean-bremen: Yes, this had been deprecated before (actually renamed to `hardlink_to`).

mrbean-bremen (commented):
Thank you for that, this is helpful! If you will exclude the extra requirement tests for Python 3.12, and add a release note (under *Infrastructure*), I will happily merge this.

browniebroke:
I'll do that tomorrow

pyfakefs/fake_filesystem.py
```diff
@@ -3446,19 +3445,7 @@ def islink(self, path: AnyPath) -> bool:
...
-
-        Returns:
-            `False` on posix systems.
-            `True` if path is a junction on Windows.
-
-        Raises:
-            TypeError: if path is None.
-        """
-        # TODO: implement junction on Windows
+        """Junction are never faked."""
```
mrbean-bremen: Typo: Junction -> Junctions. 
You may also add something like "Returns `False` - junctions are never faked."

.github/workflows/testsuite.yml
```diff
@@ -70,17 +58,17 @@ jobs:
         fi
       shell: bash
     - name: Install extra dependencies
-      if: ${{ ! matrix.experimental }}
+      if: ${{ python-version == "3.12-dev" }}
```
mrbean-bremen, line 61: Shoudln't that be != ?

browniebroke: Of course 🤦🏻

browniebroke: The string literal must be written using single quotes: https://docs.github.com/en/actions/learn-github-actions/expressions#literals

browniebroke:
Do I need to update the documentation anywhere?

mrbean-bremen (approved):
👍

mrbean-bremen:
> Do I need to update the documentation anywhere?

No, there is no need. The API documentation is automatically generated, and at this stage I wouldn't announce support for Python 3.12 yet.

browniebroke:
> The API documentation is automatically generated

Yes, I noticed that the new methods I've added aren't showing up because of the conditionals I added [here](https://github.com/pytest-dev/pyfakefs/pull/764/commits/dcdb45879c7b86f8913173bc415658a419bc4e59) since the docs aren't generated using Python 3.12. It's probably fine for now, but something to keep in mind when official support is added.

mrbean-bremen:
Good point, I wasn't aware of this.

PR #765, opened 2022-12-22
Migrate to setuptools declarative syntax

browniebroke:
#### Describe the changes
Replace `setup.py` by `setup.cfg`: https://setuptools.pypa.io/en/latest/userguide/declarative_config.html.

Generated using https://pypi.org/project/setuptools-py2cfg/

#### Tasks
- [ ] Fix or feature added
- [x] Unit tests passing


setup.cfg
```diff
@@ -1,4 +1,61 @@
 [metadata]
-description_file = README.md
```
browniebroke, line 2: There is no mention of this option in the setuptools docs. This seems to be redundant/replaced by `long_description = file:...`. According to [this Stackoverflow comment](https://stackoverflow.com/questions/60084128/does-description-file-in-setup-cfg-section-metadata-have-any-effect#comment108118088_60084977), this field was specific to PBR

mrbean-bremen: This has been there since the beginning, I never questioned it... but you are right.

While at it, maybe we should also add an `pyproject.toml` file?

setup.cfg
```diff
@@ -1,4 +1,61 @@
...
+    test
+    file
+    os
+    shutil
+    glob
+    mocking
+    unittest
+    fakes
+    filesystem
+    unit
```
mrbean-bremen, line 23: probablly remove `glob` (which is no longer mocked) and add `pathlib` instead. Also not sure if `unit` makes sense as a keyword. Ah, and add `pytest`.

setup.cfg
browniebroke: > While at it, maybe we should also add an `pyproject.toml` file?

What do you think we should put in it? Should I just set the `build-system` key?

mrbean-bremen: Yes, I think this would be sufficient for the time being - thanks!

mrbean-bremen (approved):

PR #766, opened 2022-12-23
Fix dockertests on branches containing slashes

browniebroke:
#### Describe the changes
I noticed that the dockertests [were failing](https://github.com/browniebroke/pyfakefs/actions/runs/3766893340/jobs/6403878811#step:3:977) on my branches. 

I thought it was because they are on a fork, but it's actually because my branches contained slashes `/` which means the [`$(basename $GITHUB_REF)`](https://github.com/pytest-dev/pyfakefs/blob/c617fceb91bf5e4331752fcabb9c1620859b3222/.github/workflows/dockertests.yml#L17) didn't resolve correctly to the branch name. GitHub provides another env variable `GITHUB_REF_NAME` which contains the correct value, so we can just use that instead. 

Fixing this [broke the `unzip` commands](https://github.com/browniebroke/pyfakefs/actions/runs/3766946009/jobs/6403987731#step:3:989) in the Dockerfiles, which we can avoid by setting the filename with `-O` when downloading with `wget`.

#### Tasks
- [x] Fix or feature added
- [x] Unit tests passing

### DIFF
--- a/.github/workflows/dockerfiles/Dockerfile_debian
+++ b/.github/workflows/dockerfiles/Dockerfile_debian
@@ -32,8 +32,8 @@ RUN apt-get clean
 RUN useradd -u 1000 pyfakefs
 
 RUN mkdir -p work \
-    && wget https://github.com/$github_repo/archive/$github_branch.zip \
-    && unzip $github_branch.zip -d work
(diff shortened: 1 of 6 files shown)


mrbean-bremen (approved):
Interesting. I usually don't use slashes in branch names, at least not in this repo, so I never ran into this.

PR #791, opened 2023-03-09
remove engine overwrite in pandas TextFileReader patch

satylogin:
#### Describe the changes
TextFileReader currently overwrites the engine as python when patching for pandas. I am not sure why is it needed since the default engine is "c" and "python" engine doesn't expects certain params like "float_precision" which results in it failing with the version of pandas that we have (1.1.5).

#### Tasks
- [ ] Fix or feature added
- [x] Unit tests passing

### DIFF
--- a/pyfakefs/patched_packages.py
+++ b/pyfakefs/patched_packages.py
@@ -123,6 +123,7 @@ if patch_pandas:
 
         class TextFileReader(parsers.TextFileReader):
             def __init__(self, *args, **kwargs):
+                kwargs["engine"] = "python"
                 super().__init__(*args, **kwargs)
 
         def __getattr__(self, name):
--- a/pyfakefs/patched_packages.py
+++ b/pyfakefs/patched_packages.py
@@ -32,6 +32,13 @@ try:
 except ImportError:
     locks = None
 
+# From pandas v 1.2 onwards the python fs functions are used even when the engine
+# selected is "c". This means that we don't explicitly have to change the engine.
+patch_pandas = parsers is not None and [int(v) for v in pd.__version__.split(".")] < [
+    1,
+    2,
+    0,
+]
 
 def get_modules_to_patch():
     modules_to_patch = {}
@@ -44,14 +51,14 @@ def get_modules_to_patch():
 
 def get_classes_to_patch():
     classes_to_patch = {}
-    if parsers is not None:
+    if patch_pandas:
         classes_to_patch["TextFileReader"] = "pandas.io.parsers"
     return classes_to_patch
 
 
 def get_fake_module_classes():
     fake_module_classes = {}
-    if parsers is not None:
+    if patch_pandas:
         fake_module_classes["TextFileReader"] = FakeTextFileReader
     return fake_module_classes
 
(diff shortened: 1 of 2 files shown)


mrbean-bremen:
Hm, that is strange. The patch was needed because with the "c" backend, filesystem calls cannot be patched. If I comment it out locally, the respective tests (`test_read_csv` and `test_read_table`) fail as I would have expected. 
Right now I have no idea how that works in the CI tests (the tests trigger the use of filesystem calls that have to be patched). 
So, I have to understand if the tests are not testing what they are supposed to, or if for some reason this indeed works in the CI, before I'm going to merge this (or not).

mrbean-bremen:
I'm still not quite sure what is going on, but in the CI tests the patched code is never called - it seems to go another path that uses the "python" engine by default. So in this case the change has no effect. If the code _is_ called (as in my local tests), the tests fail with the change, so I don't see what the PR would achieve.
You can always choose to [not use the patches](https://pytest-pyfakefs.readthedocs.io/en/latest/usage.html#use-known-patches) (which is what the PR practically does), and while I still want to understand what goes on in the CI tests, I'm most probably not going to merge this.

mrbean-bremen:
Ok, I now understand what is going on, and your patch makes sense, after all - just a bit changed.
The behavior of `pandas` has changed, and from version 1.2 onwards the patch is no longer needed, as internally Python fs functions are used even with the C engine. From version 1.3 onwards the patched code is no longer used, so the patch makes no difference at all, but specifically for version 1.2 it indeed makes a difference. 

Sorry for the confusion... You can adapt the PR accordingly, if you want (will add a couple of comments in a moment), or I can make the changes myself.

pyfakefs/patched_packages.py
```diff
@@ -117,7 +117,6 @@ def __init__(self, _):
 
         class TextFileReader(parsers.TextFileReader):
             def __init__(self, *args, **kwargs):
-                kwargs["engine"] = "python"
```
mrbean-bremen, line 120: The whole patching of pandas (all what is currently inside the `if parsers is not None` conditions) can be removed for pandas >= 1.2. You can do something like this for the new condition:
```py
import pandas as pd
...

patch_pandas = (
    parsers is not None and [int(v) for v in pd.__version__.split(".")] < [1, 2, 0]
)
```

Also, please add an entry to the release notes - thanks!

mrbean-bremen:
Oh, and I'm afraid that this will not help you with version 1.1.5 - this still needs the patch to work correctly. Though if you don't need that functionality, you can always switch off that patching, as mentioned before.

pyfakefs/patched_packages.py
```diff
@@ -94,7 +95,12 @@ def __getattr__(self, name):
             return getattr(self._xlrd_module, name)
 
 
-if parsers is not None:
+# From pandas v 1.2 onwards the python fs functions are used even when the engine selected is "c".
+# This means that we don't explicitly have to change the engine.
+patch_pandas = (
```
mrbean-bremen: There are two other places above with use `if parsers is not None` - these also have to be adapted to use `patch_pandas` instead.

pyfakefs/patched_packages.py
```diff
@@ -94,7 +95,14 @@ def __getattr__(self, name):
             return getattr(self._xlrd_module, name)
 
 
-if parsers is not None:
+# From pandas v 1.2 onwards the python fs functions are used even when the engine selected is "c".
+# This means that we don't explicitly have to change the engine.
+patch_pandas = parsers is not None and [int(v) for v in pd.__version__.split(".")] < [
```
mrbean-bremen, line 100: There are two other places above which use `if parsers is not None`, which also have to be adapted.

pyfakefs/patched_packages.py
satylogin: Got it, let me update.

satylogin: Updated.

mrbean-bremen (approved):
Thank you!

satylogin:
Thanks for all your help !!

PR #818, opened 2023-04-22
Re-create the temp directory on resetting the filesystem

### LINKED ISSUE #814: INTERNALERROR /path/to/file is not in the subpath of '\\/path/to/file'
**Describe the bug**

In unit testing, I need to emulate a Windows OS.  My base machine is macOS.

When using `pyfakefs.fake_filesystem.OSType.WINDOWS` + `tempfile` + `--log-cli-level=DEBUG`, I get the below internal error.
 
**How To Reproduce**

```python
import tempfile

import pyfakefs.fake_filesystem


def test_foo(fs) -> None:
    fs.os = pyfakefs.fake_filesystem.OSType.WINDOWS

    with tempfile.TemporaryDirectory():
        pass
```

Outputs the following:

```python
> pytest test/repo/test/test_play.py -s --log-cli-level=DEBUG
=============================================================================================================== test session starts ===============================================================================================================
platform darwin -- Python 3.10.9, pytest-7.2.2, pluggy-1.0.0
rootdir: /Users/user/code/repo/test, configfile: pyproject.toml
plugins: pyfakefs-5.2.2, mock-3.10.0, subtests-0.10.0
collected 1 item

test/repo/test/test_play.py::test_foo
INTERNALERROR> Traceback (most recent call last):
INTERNALERROR>   File "/Users/user/code/repo/venv/lib/python3.10/site-packages/_pytest/main.py", line 270, in wrap_session
INTERNALERROR>     session.exitstatus = doit(config, session) or 0
INTERNALERROR>   File "/Users/user/code/repo/venv/lib/python3.10/site-packages/_pytest/main.py", line 324, in _main
INTERNALERROR>     config.hook.pytest_runtestloop(session=session)
INTERNALERROR>   File "/Users/user/code/repo/venv/lib/python3.10/site-packages/pluggy/_hooks.py", line 265, in __call__
INTERNALERROR>     return self._hookexec(self.name, self.get_hookimpls(), kwargs, firstresult)
INTERNALERROR>   File "/Users/user/code/repo/venv/lib/python3.10/site-packages/pluggy/_manager.py", line 80, in _hookexec
INTERNALERROR>     return self._inner_hookexec(hook_name, methods, kwargs, firstresult)
INTERNALERROR>   File "/Users/user/code/repo/venv/lib/python3.10/site-packages/pluggy/_callers.py", line 60, in _multicall
INTERNALERROR>     return outcome.get_result()
INTERNALERROR>   File "/Users/user/code/repo/venv/lib/python3.10/site-packages/pluggy/_result.py", line 60, in get_result
INTERNALERROR>     raise ex[1].with_traceback(ex[2])
INTERNALERROR>   File "/Users/user/code/repo/venv/lib/python3.10/site-packages/pluggy/_callers.py", line 39, in _multicall
INTERNALERROR>     res = hook_impl.function(*args)
INTERNALERROR>   File "/Users/user/code/repo/venv/lib/python3.10/site-packages/_pytest/main.py", line 349, in pytest_runtestloop
INTERNALERROR>     item.config.hook.pytest_runtest_protocol(item=item, nextitem=nextitem)
INTERNALERROR>   File "/Users/user/code/repo/venv/lib/python3.10/site-packages/pluggy/_hooks.py", line 265, in __call__
INTERNALERROR>     return self._hookexec(self.name, self.get_hookimpls(), kwargs, firstresult)
INTERNALERROR>   File "/Users/user/code/repo/venv/lib/python3.10/site-packages/pluggy/_manager.py", line 80, in _hookexec
INTERNALERROR>     return self._inner_hookexec(hook_name, methods, kwargs, firstresult)
INTERNALERROR>   File "/Users/user/code/repo/venv/lib/python3.10/site-packages/pluggy/_callers.py", line 60, in _multicall
INTERNALERROR>     return outcome.get_result()
INTERNALERROR>   File "/Users/user/code/repo/venv/lib/python3.10/site-packages/pluggy/_result.py", line 60, in get_result
INTERNALERROR>     raise ex[1].with_traceback(ex[2])
INTERNALERROR>   File "/Users/user/code/repo/venv/lib/python3.10/site-packages/pluggy/_callers.py", line 39, in _multicall
INTERNALERROR>     res = hook_impl.function(*args)
INTERNALERROR>   File "/Users/user/code/repo/venv/lib/python3.10/site-packages/_pytest/runner.py", line 112, in pytest_runtest_protocol
INTERNALERROR>     runtestprotocol(item, nextitem=nextitem)
INTERNALERROR>   File "/Users/user/code/repo/venv/lib/python3.10/site-packages/_pytest/runner.py", line 131, in runtestprotocol
INTERNALERROR>     reports.append(call_and_report(item, "call", log))
INTERNALERROR>   File "/Users/user/code/repo/venv/lib/python3.10/site-packages/_pytest/runner.py", line 224, in call_and_report
INTERNALERROR>     hook.pytest_runtest_logreport(report=report)
INTERNALERROR>   File "/Users/user/code/repo/venv/lib/python3.10/site-packages/pluggy/_hooks.py", line 265, in __call__
INTERNALERROR>     return self._hookexec(self.name, self.get_hookimpls(), kwargs, firstresult)
INTERNALERROR>   File "/Users/user/code/repo/venv/lib/python3.10/site-packages/pluggy/_manager.py", line 80, in _hookexec
INTERNALERROR>     return self._inner_hookexec(hook_name, methods, kwargs, firstresult)
INTERNALERROR>   File "/Users/user/code/repo/venv/lib/python3.10/site-packages/pluggy/_callers.py", line 60, in _multicall
INTERNALERROR>     return outcome.get_result()
INTERNALERROR>   File "/Users/user/code/repo/venv/lib/python3.10/site-packages/pluggy/_result.py", line 60, in get_result
INTERNALERROR>     raise ex[1].with_traceback(ex[2])
INTERNALERROR>   File "/Users/user/code/repo/venv/lib/python3.10/site-packages/pluggy/_callers.py", line 39, in _multicall
INTERNALERROR>     res = hook_impl.function(*args)
INTERNALERROR>   File "/Users/user/code/repo/venv/lib/python3.10/site-packages/_pytest/terminal.py", line 551, in pytest_runtest_logreport
INTERNALERROR>     line = self._locationline(rep.nodeid, *rep.location)
INTERNALERROR>   File "/Users/user/code/repo/venv/lib/python3.10/site-packages/_pytest/terminal.py", line 878, in _locationline
INTERNALERROR>     res = mkrel(nodeid)
INTERNALERROR>   File "/Users/user/code/repo/venv/lib/python3.10/site-packages/_pytest/terminal.py", line 868, in mkrel
INTERNALERROR>     line = self.config.cwd_relative_nodeid(nodeid)
INTERNALERROR>   File "/Users/user/code/repo/venv/lib/python3.10/site-packages/_pytest/config/__init__.py", line 1104, in cwd_relative_nodeid
INTERNALERROR>     nodeid = bestrelpath(self.invocation_params.dir, fullpath)
INTERNALERROR>   File "/Users/user/code/repo/venv/lib/python3.10/site-packages/_pytest/pathlib.py", line 710, in bestrelpath
INTERNALERROR>     reldirectory = directory.relative_to(base)
INTERNALERROR>   File "/Users/user/.pyenv/versions/3.10.9/lib/python3.10/pathlib.py", line 818, in relative_to
INTERNALERROR>     raise ValueError("{!r} is not in the subpath of {!r}"
INTERNALERROR> ValueError: '/Users/user/code/repo' is not in the subpath of '\\/Users/user/code/repo' OR one path is relative and the other is absolute.
  File "/Users/user/code/repo/venv/lib/python3.10/site-packages/pyfakefs/fake_filesystem.py", line 1618, in get_object_from_normpath
    target = target.get_entry(component)  # type: ignore
  File "/Users/user/code/repo/venv/lib/python3.10/site-packages/pyfakefs/fake_file.py", line 542, in get_entry
    return self.entries[to_string(pathname_name)]
KeyError: 'Users'

During handling of the above exception, another exception occurred:

  File "/Users/user/code/repo/venv/bin/pytest", line 8, in <module>
    sys.exit(console_main())
  File "/Users/user/code/repo/venv/lib/python3.10/site-packages/_pytest/config/__init__.py", line 190, in console_main
    code = main()
  File "/Users/user/code/repo/venv/lib/python3.10/site-packages/_pytest/config/__init__.py", line 167, in main
    ret: Union[ExitCode, int] = config.hook.pytest_cmdline_main(
  File "/Users/user/code/repo/venv/lib/python3.10/site-packages/pluggy/_hooks.py", line 265, in __call__
    return self._hookexec(self.name, self.get_hookimpls(), kwargs, firstresult)
  File "/Users/user/code/repo/venv/lib/python3.10/site-packages/pluggy/_manager.py", line 80, in _hookexec
    return self._inner_hookexec(hook_name, methods, kwargs, firstresult)
  File "/Users/user/code/repo/venv/lib/python3.10/site-packages/pluggy/_callers.py", line 60, in _multicall
    return outcome.get_result()
  File "/Users/user/code/repo/venv/lib/python3.10/site-packages/pluggy/_result.py", line 60, in get_result
    raise ex[1].with_traceback(ex[2])
  File "/Users/user/code/repo/venv/lib/python3.10/site-packages/pluggy/_callers.py", line 39, in _multicall
    res = hook_impl.function(*args)
  File "/Users/user/code/repo/venv/lib/python3.10/site-packages/_pytest/main.py", line 317, in pytest_cmdline_main
    return wrap_session(config, _main)
  File "/Users/user/code/repo/venv/lib/python3.10/site-packages/_pytest/main.py", line 302, in wrap_session
    os.chdir(session.startpath)
  File "/Users/user/code/repo/venv/lib/python3.10/site-packages/pyfakefs/fake_os.py", line 1354, in wrapped
    return f(*args, **kwargs)
  File "/Users/user/code/repo/venv/lib/python3.10/site-packages/pyfakefs/fake_os.py", line 421, in chdir
    self.filesystem.confirmdir(path)
  File "/Users/user/code/repo/venv/lib/python3.10/site-packages/pyfakefs/fake_filesystem.py", line 2810, in confirmdir
    self.resolve(target_directory, check_owner=check_owner),
  File "/Users/user/code/repo/venv/lib/python3.10/site-packages/pyfakefs/fake_filesystem.py", line 1692, in resolve
    return self.get_object_from_normpath(
  File "/Users/user/code/repo/venv/lib/python3.10/site-packages/pyfakefs/fake_filesystem.py", line 1627, in get_object_from_normpath
    self.raise_os_error(errno.ENOENT, path)
  File "/Users/user/code/repo/venv/lib/python3.10/site-packages/pyfakefs/fake_filesystem.py", line 418, in raise_os_error
    raise OSError(err_no, message, filename)
FileNotFoundError: [Errno 2] No such file or directory in the fake filesystem: 'C:\\Users\\user\\code\\repo'
```

**Your environment**

I have `pyfakefs==5.2.2`.  Note the current version printing from the issue template is broken.

```bash
> python -c "import platform; print(platform.platform())"
macOS-12.6-arm64-arm-64bit
> python -c "import sys; print('Python', sys.version)"
Python 3.10.9 (main, Jan  6 2023, 11:19:20) [Clang 14.0.0 (clang-1400.0.29.102)]
> python -c "from pyfakefs.fake_filesystem import __version__; print('pyfakefs', __version__)"
  File "<string>", line 1, in <module>
ImportError: cannot import name '__version__' from 'pyfakefs.fake_filesystem' (/Users/user/proj/proj_venv/lib/python3.10/site-packages/pyfakefs/fake_filesystem.py)
> python -c "import pytest; print('pytest', pytest.__version__)"
pytest 7.2.2
```

mrbean-bremen:
- the temp dir is now created during filesystem initialization instead of at patcher setup
- fixes #814

This slightly changes the behavior if you are creating a fake filesystem without using any of the recommended methods (e.g. the temp directory is now also created in this case), but I think that this is ok.

#### Tasks
- [x] Fix or feature added
- [x] Unit tests passing

### DIFF
--- a/pyfakefs/fake_filesystem.py
+++ b/pyfakefs/fake_filesystem.py
@@ -235,7 +235,7 @@ class FakeFilesystem:
         # file systems on non-case-sensitive systems and vice verse
         self.is_case_sensitive: bool = not (self.is_windows_fs or self._is_macos)
 
-        self.root = FakeDirectory(self.path_separator, filesystem=self)
+        self.root: FakeDirectory
         self._cwd = ""
 
         # We can't query the current value without changing it:
@@ -251,10 +251,9 @@ class FakeFilesystem:
         self.last_ino: int = 0
         self.last_dev: int = 0
         self.mount_points: Dict[AnyString, Dict] = OrderedDict()
-        self._add_root_mount_point(total_size)
-        self._add_standard_streams()
-        self.dev_null: Any = FakeNullFile(self)
-        self._create_temp_dir()
+        self.dev_null: Any = None
+        self.reset(total_size=total_size, init_pathlib=False)
+
         # set from outside if needed
         self.patch_open_code = PatchMode.OFF
         self.shuffle_listdir_results = False
@@ -338,22 +337,24 @@ class FakeFilesystem:
         self.reset()
         FakePathModule.reset(self)
 
-    def reset(self, total_size: Optional[int] = None):
+    def reset(self, total_size: Optional[int] = None, init_pathlib: bool = True):
         """Remove all file system contents and reset the root."""
         self.root = FakeDirectory(self.path_separator, filesystem=self)
 
         self.dev_null = FakeNullFile(self)
-        self.open_files = []
-        self._free_fd_heap = []
+        self.open_files.clear()
+        self._free_fd_heap.clear()
         self.last_ino = 0
         self.last_dev = 0
-        self.mount_points = OrderedDict()
+        self.mount_points.clear()
         self._add_root_mount_point(total_size)
         self._add_standard_streams()
-        self._create_temp_dir()
-        from pyfakefs import fake_pathlib
+        if self.create_temp_dir:
+            self._create_temp_dir()
+        if init_pathlib:
+            from pyfakefs import fake_pathlib
 
-        fake_pathlib.init_module(self)
+            fake_pathlib.init_module(self)
 
     def _add_root_mount_point(self, total_size):
         mount_point = "C:" if self.is_windows_fs else self.path_separator
@@ -2925,8 +2926,6 @@ class FakeFilesystem:
         self._add_open_file(StandardStreamWrapper(sys.stderr))
 
     def _create_temp_dir(self):
-        if not self.create_temp_dir:
-            return
         # the temp directory is assumed to exist at least in `tempfile`,
         # so we create it here for convenience
         temp_dir = tempfile.gettempdir()
--- a/pyfakefs/fake_filesystem.py
+++ b/pyfakefs/fake_filesystem.py
@@ -85,6 +85,7 @@ import heapq
 import os
 import random
 import sys
+import tempfile
 from collections import namedtuple, OrderedDict
 from doctest import TestResults
 from enum import Enum
@@ -197,12 +198,17 @@ class FakeFilesystem:
         path_separator: str = os.path.sep,
         total_size: Optional[int] = None,
         patcher: Any = None,
+        create_temp_dir: bool = False,
     ) -> None:
         """
         Args:
             path_separator:  optional substitute for os.path.sep
             total_size: if not None, the total size in bytes of the
                 root filesystem.
+            patcher: the Patcher instance if created from the Patcher
+            create_temp_dir: If True, a temp directory is created on initialization.
+                Under Posix, if the temp directory is not `/tmp`, a link to the temp
+                path is additionally created at `/tmp`.
 
         Example usage to use the same path separator under all systems:
 
@@ -212,6 +218,7 @@ class FakeFilesystem:
         self.path_separator: str = path_separator
         self.alternative_path_separator: Optional[str] = os.path.altsep
         self.patcher = patcher
+        self.create_temp_dir = create_temp_dir
         if path_separator != os.sep:
             self.alternative_path_separator = None
 
@@ -247,6 +254,7 @@ class FakeFilesystem:
         self._add_root_mount_point(total_size)
         self._add_standard_streams()
         self.dev_null: Any = FakeNullFile(self)
+        self._create_temp_dir()
         # set from outside if needed
         self.patch_open_code = PatchMode.OFF
         self.shuffle_listdir_results = False
@@ -342,6 +350,7 @@ class FakeFilesystem:
         self.mount_points = OrderedDict()
         self._add_root_mount_point(total_size)
         self._add_standard_streams()
+        self._create_temp_dir()
         from pyfakefs import fake_pathlib
 
         fake_pathlib.init_module(self)
@@ -1380,7 +1389,7 @@ class FakeFilesystem:
         """
         if check_link and self.islink(file_path):
             return True
-        path = to_string(make_string_path(file_path))
+        path = to_string(self.make_string_path(file_path))
         if path is None:
             raise TypeError
         if not path:
@@ -2914,6 +2923,21 @@ class FakeFilesystem:
         self._add_open_file(StandardStreamWrapper(sys.stdin))
         self._add_open_file(StandardStreamWrapper(sys.stdout))
         self._add_open_file(StandardStreamWrapper(sys.stderr))
+
+    def _create_temp_dir(self):
+        if not self.create_temp_dir:
+            return
+        # the temp directory is assumed to exist at least in `tempfile`,
+        # so we create it here for convenience
+        temp_dir = tempfile.gettempdir()
+        if not self.exists(temp_dir):
+            self.create_dir(temp_dir)
+        if sys.platform != "win32" and not self.exists("/tmp"):
+            # under Posix, we also create a link in /tmp if the path does not exist
+            self.create_symlink("/tmp", temp_dir)
+            # reset the used size to 0 to avoid having the link size counted
+            # which would make disk size tests more complicated
+            next(iter(self.mount_points.values()))["used_size"] = 0
 
 
 def _run_doctest() -> TestResults:
--- a/CHANGES.md
+++ b/CHANGES.md
@@ -1,6 +1,12 @@
 # pyfakefs Release Notes
 The released versions correspond to PyPI releases.
 
+## Unreleased
+
+### Fixes
+* Re-create temp directory if it had been created before on resetting file system
+  (see [#814](../../issues/814)).
+
 ## [Version 5.2.2](https://pypi.python.org/pypi/pyfakefs/5.2.2) (2023-04-13)
 Fixes a regression in 5.2.0
 
--- a/pyfakefs/fake_filesystem_unittest.py
+++ b/pyfakefs/fake_filesystem_unittest.py
@@ -35,7 +35,7 @@ Existing unit tests that use the real file system can be retrofitted to use
 pyfakefs by simply changing their base class from `:py:class`unittest.TestCase`
 to `:py:class`pyfakefs.fake_filesystem_unittest.TestCase`.
 """
-import _io  # type:ignore [import]
+import _io  # type:ignore[import]
 import doctest
 import functools
 import genericpath
@@ -844,7 +844,7 @@ class Patcher:
             self._stubs.smart_unset_all()
         self._stubs = mox3_stubout.StubOutForTesting()
 
-        self.fs = fake_filesystem.FakeFilesystem(patcher=self)
+        self.fs = fake_filesystem.FakeFilesystem(patcher=self, create_temp_dir=True)
         self.fs.patch_open_code = self.patch_open_code
         for name in self._fake_module_classes:
             self.fake_modules[name] = self._fake_module_classes[name](self.fs)
@@ -876,7 +876,6 @@ class Patcher:
         if self.has_fcopy_file:
             shutil._HAS_FCOPYFILE = False  # type: ignore[attr-defined]
 
-        temp_dir = tempfile.gettempdir()
         with warnings.catch_warnings():
             # ignore warnings, see #542 and #614
             warnings.filterwarnings("ignore")
@@ -890,17 +889,6 @@ class Patcher:
         self.start_patching()
         linecache.open = self.original_open  # type: ignore[attr-defined]
         tokenize._builtin_open = self.original_open  # type: ignore
-
-        # the temp directory is assumed to exist at least in `tempfile`,
-        # so we create it here for convenience
-        assert self.fs is not None
-        self.fs.create_dir(temp_dir)
-        if sys.platform != "win32" and not self.fs.exists("/tmp"):
-            # under Posix, we also create a link in /tmp if the path does not exist
-            self.fs.create_symlink("/tmp", temp_dir)
-            # reset the used size to 0 to avoid having the link size counted
-            # which would make disk size tests more complicated
-            next(iter(self.fs.mount_points.values()))["used_size"] = 0
 
     def start_patching(self) -> None:
         if not self._patching:
--- a/pyfakefs/pytest_tests/pytest_plugin_test.py
+++ b/pyfakefs/pytest_tests/pytest_plugin_test.py
@@ -2,6 +2,7 @@
 import os
 import tempfile
 
+from pyfakefs.fake_filesystem import OSType
 from pyfakefs.fake_filesystem_unittest import Pause
 import pyfakefs.pytest_tests.io
 
@@ -60,3 +61,18 @@ def test_use_own_io_module(fs):
 
     stream = pyfakefs.pytest_tests.io.InputStream(filepath)
     assert stream.read() == "bar"
+
+
+def test_switch_to_windows(fs):
+    fs.os = OSType.WINDOWS
+    assert os.path.exists(tempfile.gettempdir())
+
+
+def test_switch_to_linux(fs):
+    fs.os = OSType.LINUX
+    assert os.path.exists(tempfile.gettempdir())
+
+
+def test_switch_to_macos(fs):
+    fs.os = OSType.MACOS
+    assert os.path.exists(tempfile.gettempdir())
--- a/pyfakefs/tests/fake_filesystem_unittest_test.py
+++ b/pyfakefs/tests/fake_filesystem_unittest_test.py
@@ -657,11 +657,11 @@ class TestTempDirCreation(fake_filesystem_unittest.TestCase):
     def setUp(self):
         self.setUpPyfakefs()
 
-    def testTempDirExists(self):
+    def test_tempdir_exists(self):
         self.assertTrue(os.path.exists(tempfile.gettempdir()))
 
     @unittest.skipIf(sys.platform == "win32", "POSIX only test")
-    def testTmpExists(self):
+    def test_tmp_exists(self):
         # directory or link under Linux, link under macOS
         self.assertTrue(os.path.exists("/tmp"))
 


davidlbaird:
A large number of test modules are not using the patcher to instantiate fake_filesystem and may be mocking or making use of tempdir independently of pyfakefs. Please be careful with this change.

pyfakefs/fake_filesystem.py
```diff
@@ -247,6 +248,7 @@ def __init__(
         self._add_root_mount_point(total_size)
         self._add_standard_streams()
         self.dev_null: Any = FakeNullFile(self)
+        self._create_temp_dir()
```
davidlbaird, line 251: Perhaps create a temp dir only if self.patcher is set?

pyfakefs/fake_filesystem.py
```diff
@@ -2915,6 +2918,19 @@ def _add_standard_streams(self) -> None:
...
 
+    def _create_temp_dir(self):
+        # the temp directory is assumed to exist at least in `tempfile`,
+        # so we create it here for convenience
+        temp_dir = tempfile.gettempdir()
+        if not self.exists(temp_dir):
+            self.create_dir(temp_dir)
+        if sys.platform != "win32" and not self.exists("/tmp"):
+            # under Posix, we also create a link in /tmp if the path does not exist
+            self.create_symlink("/tmp", temp_dir)
```
davidlbaird, line 2929: See https://docs.python.org/3/library/tempfile.html#tempfile.gettempdir
The possible values of tempfile.gettempdir() would be '/tmp', '/var/tmp', or '/usr/tmp' and I would think that if either '/var/tmp' or '/usr/tmp' were created, then a symlink '/tmp' would not be part of the native file system.

pyfakefs/fake_filesystem.py
davidlbaird, line 251: This is an unexpected behavior change to pyfakefs. I found a lot of unit tests that create a /tmp directory in their fake filesystem object, and they will fail on a new pyfakefs that already creates one in the fake filesystem.

pyfakefs/tests/fake_filesystem_test.py
```diff
@@ -852,6 +853,9 @@ def tearDown(self):
 
     def test_create_top_level_directory(self):
         top_level_dir = "/x"
+        # remove the existing temp dir
```
davidlbaird, line 856: Is this because /tmp is not expected further down? This seems to be forcing the test function to conform to the new behavior.

mrbean-bremen: True.
First - thanks for the review. It is good to have someone else to check the code again...

For a bit of background: 
Previously (that was several years ago), the `tempfile` module was patched in `pyfakefs`, but the patch was incomplete and had problems, and broke with every new Python version. So I removed it from patching (together with `glob` which had a similar problem) and solely relied on the patched `os` / `os.path` functionality. That worked, except that `tempfile` assumes an existing temp directory, otherwise it crashes, so I decided to create that directory on startup.

It is true that this changes the behavior [if you expected an empty filesystem](https://pytest-pyfakefs.readthedocs.io/en/latest/troubleshooting.html#os-temporary-directories), but so far I had no reports that this causes big problems.

I was also not sure about the usage pattern of `pyfakefs`. Recently I did some statistics with the repositories shown as dependent on `pyfakefs` (as this is the only source I have for that), counting only repositories that had at least 100 stars (about 50, so not really a large sample), and found that of these, more than 40% used it via `pytest`, and about the same number via the `unittest` support added by @jmcgeheeiv. Another 5% used the patcher directly, and the rest (4 in this case) had been manually patching as originally used by Google.
If assume that the usage you are referring to is inside Google, which is good to know. 

More to the point: I could add an argument to the fake filesystem initialization (defaulting to `False`) that would control if the temp file shall be created. I would probably set that to `True` if initialized from the patcher - this way, it would retain the current behavior if invoking `pyfakefs` via `pytest` or `unittest`, and would also retain the behavior if creating it manually. Would that be ok for you?

mrbean-bremen: Generally, I think that we have to keep 2 usage scenarios in mind: replacing the whole fs using the patcher (directly or via the `fs` fixture, `fake_filesystem_unittest.TestCase` or `@patchfs`), and the usage by directly instantiating the fake filesystem and the fake modules, where needed, possibly subclassing them. The first use case is what I mostly had in mind so far, and that is mostly used by people I get feedback from. The second scenario is the one originally used (and obviously still used) at Google and certainly elsewhere, but so far we had no feedback from these users.

davidlbaird: Yes, there were previously three additional modules that were removed, `fake_tempfile`, `fake_filesystem_glob` and `fake_filesystem_shutil`. When mocking modules with `fake_filesystem` through `unittest.mock`, which is common in the Google unit tests, the unit tests would add mocking behavior to `tempfile` through the `fake_tempfile` module. The other two modules are simple to mock, `shutil` and `glob` by mocking their `os` imports directly. However, `tempfile` is, as you note, very complex, and so after `fake_tempfile` was removed, those unit tests ported their implementations to `fake_filesystem_unittest` which provides a complete implementation of a POSIX file system.

In the case of adding `fake_filesyste_unittest` to unit tests, I think it is very appropriate to add the `tempfile.gettempdir` directory for POSIX file system support. I also think that resetting the file system for each test case makes a lot of sense. In fact, I encourage use of the `@fake_filesystem_unittest.patchfs` decorator instead of subclassing `fake_filessytem_unittest.TestCase`.

I think that for `fake_filesystem` however, the only expectation is a clean fake filesystem, and nothing more. In fact, internally at Google, we have other file systems other than the local filesystem that we provide fakes for, that extend pyfakefs fake filesystem, and for which there are no devices (no /dev/null that was added at some point to pyfakefs) and no /tmp directory, because they don't make sense in those situations.

It is very simple to add a temporary directory when setting up a test function to a test module that imports `fake_filesystem` and mocks the imports of modules under test, so I would prefer preserving that expectation.

I may need to add documentation on usage of `fake_filesystem` with the `unittest.mock` module, and also, how to extend `fake_filesystem` for non-POSIX, Windows or MacOS file systems, that are not patching imports of the modules under test, and are rather, accessing those file systems through other APIs.

Thanks.

mrbean-bremen: Thanks, this all makes sense. As I wrote, that is a different use case, that I hadn't really looked upon - I myself started using `pyfakefs` in a django environment under Windows (don't ask), and had always used `fake_filesystem_unittest.TestCase`, or later the `fs` fixture. The first thing I added was better Windows support, and a couple of features (like support for file system size) that I needed myself. Afterwards the development was mostly driven by issues and feature requests, and by new Python versions (other than infrastructure stuff like documentation, CI and more tests), and none of these so far was related to the usage with `unittest.mock` (it isn't even in the documentation). Hopefully this will now change...

As for this PR - I understand it is ok for you in the current incarnation (e.g. with the previous behavior in place)? I'll wait for your approval before merging.

davidlbaird: Is it possible to address the regression when resetting the `fs` fixture without adding tempfile support to the `FakeFilsystem` class by default? I understand the intent of this code change, but since it is responding to a regression of `fs` fixtures, I think it should be limited to resetting `fs` fixtures.

This change code change will impact all clients of `FileSystem`. I'm pretty sure that many of the unit tests I support (20,000 at latest count) will fail because of assumptions about the file system not containing a `/tmp` directory. The change is well intentioned, and perhaps a different implementation could provide an initial directory structure to `FakeFilesystem` through a new `__init__` arg. This could be useful for many other clients of `FakeFilesystem` with their own initial file directory structure.

In terms of `pyfakefs` design, I'd like to review recent changes, and as I said, add explanations of using the `FakeFilesystem` class for other file system interfaces not supported by the `fs` fixture. As I said, it is used by special file systems. I worked on the capability to fake a Windows file system from a unit test running on Linux many many years ago (2012?), and I am glad that in general, it is a useful feature and has continued support.

mrbean-bremen: As I wrote, after the change I made after your comment, tempfile creation is switched off by default. It is only switched on if created from the patcher (or if you explicitely switch it on), so the behavior should not change for the existing clients. Maybe you hadn't noticed the commit, as I just amended the old one...
Leaving the code in the patcher would make it difficult to call it on filesystem reset.

About the review: I will be glad if you can review the code and suggest changes. I understand you want to add the explanations to the `pyfakefs` documentation and not to your internal documentation? That would indeed by helpful!

davidlbaird: Oh... sorry... I didn't see the update in the source files. My bad. I'll continue my code review there.

pyfakefs/fake_filesystem.py
davidlbaird: I still don't understand why this added symlink is required.

pyfakefs/fake_filesystem.py
```diff
@@ -212,6 +218,7 @@ def __init__(
         self.path_separator: str = path_separator
         self.alternative_path_separator: Optional[str] = os.path.altsep
         self.patcher = patcher
+        self.create_temp_dir = create_temp_dir
```
davidlbaird, line 221: This is called twice in`__init__`, here and at line 257. Why is does the temp dir need to be created twice?

pyfakefs/fake_filesystem.py
```diff
@@ -247,6 +254,7 @@ def __init__(
         self._add_root_mount_point(total_size)
         self._add_standard_streams()
         self.dev_null: Any = FakeNullFile(self)
+        self._create_temp_dir()
```
davidlbaird, line 257: Perhaps call self.reset to consolidate all these file system updates in one method?

pyfakefs/fake_filesystem.py
davidlbaird, line 221: This is called twice in__init__, here and at line 257. Why is does the temp dir need to be created twice?

pyfakefs/fake_filesystem.py
```diff
@@ -2915,6 +2924,21 @@ def _add_standard_streams(self) -> None:
         self._add_open_file(StandardStreamWrapper(sys.stdout))
         self._add_open_file(StandardStreamWrapper(sys.stderr))
 
+    def _create_temp_dir(self):
+        if not self.create_temp_dir:
```
davidlbaird, line 2928: I would guard `_create_temp_dir` from the `reset` method instead, especially if `reset` can be called from `__init__`.

pyfakefs/fake_filesystem.py
```diff
@@ -2915,6 +2924,21 @@ def _add_standard_streams(self) -> None:
...
+            return
+        # the temp directory is assumed to exist at least in `tempfile`,
+        # so we create it here for convenience
+        temp_dir = tempfile.gettempdir()
+        if not self.exists(temp_dir):
+            self.create_dir(temp_dir)
+        if sys.platform != "win32" and not self.exists("/tmp"):
+            # under Posix, we also create a link in /tmp if the path does not exist
+            self.create_symlink("/tmp", temp_dir)
```
davidlbaird, line 2937: I still don't understand why this added symlink is required.

pyfakefs/fake_filesystem.py
mrbean-bremen: Actually no - I'm just setting the attribute here, while on ine 257 I call the method (maybe the naming could be better). The attribute is needed to know if the temp dir shall be recreated in `reset`, and I check for the attribute in the method `_create_temp_dir` to make sure it is only recreated if it had been created before.

pyfakefs/fake_filesystem.py
mrbean-bremen: You mean from `__init__`? I think I did this before, but as far as I remember the static checker did not like that the member variables have been initialized outside of `__init__`. But you are right, that is too much duplication...

pyfakefs/fake_filesystem.py
mrbean-bremen: Yes - I decided to use this one, because otherwise I would have to guard it twice. But if I call `reset` from `__init__` instead, this won't be true anymore.

pyfakefs/fake_filesystem.py
mrbean-bremen: Well, the issue author mentioned the [FHS](https://refspecs.linuxfoundation.org/FHS_3.0/fhs/ch03s18.html), which was good enough for me, though I may be wrong.

davidlbaird (approved):
Thank you. I'll need to figure out how to get notified of pull request review updates.

davidlbaird (approved):
Looks nice!

PR #824, opened 2023-05-02
Add pytype CI and pytype ignore comments where needed

mrbean-bremen:
This is just to get some starting point for pytype support.
I fixed very few problems, the rest is just ignore comments to silence the pytype errors.
I also excluded the tests (as we do for mypy), as they have a lot of problems, and at this stage, I'm not really interested in fixing these.

#### Tasks
- [x] Fix or feature added
- [x] Unit tests passing

### DIFF
--- a/pyfakefs/fake_file.py
+++ b/pyfakefs/fake_file.py
@@ -352,7 +352,7 @@ class FakeFile:
     @property
     def path(self) -> AnyStr:
         """Return the full path of the current object."""
-        names: List[AnyStr] = []
+        names: List[AnyStr] = []  # pytype: disable=invalid-annotation
         obj: Optional[FakeFile] = self
         while obj:
             names.insert(0, matching_string(self.name, obj.name))  # type: ignore
@@ -1280,7 +1280,7 @@ class FakePipeWrapper:
     def read(self, numBytes: int = -1) -> bytes:
         """Read from the real pipe."""
         if self.real_file:
-            return self.real_file.read(numBytes)
+            return self.real_file.read(numBytes)  # pytype: disable=bad-return-type
         return os.read(self.fd, numBytes)
 
     def flush(self) -> None:
--- a/pyfakefs/fake_filesystem_unittest.py
+++ b/pyfakefs/fake_filesystem_unittest.py
@@ -909,7 +909,9 @@ class Patcher:
         for (name, ft_name, ft_mod), modules in self.FS_FUNCTIONS.items():
             method, mod_name = self._fake_module_functions[ft_name][ft_mod]
             fake_module = self.fake_modules[mod_name]
-            attr = method.__get__(fake_module, fake_module.__class__)
+            attr = method.__get__(
+                fake_module, fake_module.__class__
+            )  # pytype: disable=attribute-error
             for module in modules:
                 self._stubs.smart_set(module, name, attr)
 
@@ -927,7 +929,9 @@ class Patcher:
         for fct, idx, ft in self.FS_DEFARGS:
             method, mod_name = self._fake_module_functions[ft.__name__][ft.__module__]
             fake_module = self.fake_modules[mod_name]
-            attr = method.__get__(fake_module, fake_module.__class__)
+            attr = method.__get__(
+                fake_module, fake_module.__class__
+            )  # pytype: disable=attribute-error
             new_defaults = []
             assert fct.__defaults__ is not None
             for i, d in enumerate(fct.__defaults__):
--- a/pyfakefs/fake_filesystem_unittest.py
+++ b/pyfakefs/fake_filesystem_unittest.py
@@ -78,11 +78,7 @@ from pyfakefs.fake_filesystem import (
 from pyfakefs.helpers import IS_PYPY
 from pyfakefs.mox3_stubout import StubOutForTesting
 
-try:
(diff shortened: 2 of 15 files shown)


pyfakefs/fake_file.py
```diff
@@ -352,7 +352,7 @@ def size(self, st_size: int) -> None:
     @property
     def path(self) -> AnyStr:
         """Return the full path of the current object."""
-        names: List[AnyStr] = []
+        names: List[AnyStr] = []  # pytype: disable=invalid-annotation
```
mrbean-bremen, line 355: Most of the problems are related to `AnyStr`, which seems to behave differently with `pytype` compared to `mypy` - still have to figure that out.

pyfakefs/fake_file.py
```diff
@@ -1280,7 +1280,7 @@ def fileno(self) -> int:
     def read(self, numBytes: int = -1) -> bytes:
         """Read from the real pipe."""
         if self.real_file:
-            return self.real_file.read(numBytes)
+            return self.real_file.read(numBytes)  # pytype: disable=bad-return-type
```
mrbean-bremen, line 1283: also related to `AnyStr`

pyfakefs/fake_filesystem_unittest.py
```diff
@@ -909,7 +907,9 @@ def patch_functions(self) -> None:
         for (name, ft_name, ft_mod), modules in self.FS_FUNCTIONS.items():
             method, mod_name = self._fake_module_functions[ft_name][ft_mod]
             fake_module = self.fake_modules[mod_name]
-            attr = method.__get__(fake_module, fake_module.__class__)
+            attr = method.__get__(
+                fake_module, fake_module.__class__
+            )  # pytype: disable=attribute-error
```
mrbean-bremen, line 912: I think these these are unavoidable due to the dynamic nature of the code

pyfakefs/fake_io.py
```diff
@@ -102,7 +102,7 @@ def open(
                 for sn in self.skip_names
             ]
         ):
-            return io.open(
+            return io.open(  # pytype: disable=wrong-arg-count
```
mrbean-bremen, line 105: I have no idea where this comes from from - for some reason, it doesn't expect the last argument (`opener`).

pyfakefs/fake_path.py
```diff
@@ -501,7 +500,7 @@ def __init__(self, filesystem: "FakeFilesystem"):
             Args:
                 filesystem: FakeFilesystem used to provide file system information
             """
-            import nt
+            import nt  # type:ignore[import]
```
mrbean-bremen, line 503: `pytype` does not see that this is in a specific platform branch - maybe we have to tell it some other way...

pyfakefs/fake_pathlib.py
```diff
@@ -521,18 +521,18 @@ def __new__(cls, *args, **kwargs):
                 else FakePathlibModule.PosixPath
             )
         if sys.version_info < (3, 12):
-            return cls._from_parts(args)
+            return cls._from_parts(args)  # pytype: disable=attribute-error
```
mrbean-bremen, line 524: `pathlib` has changed changed the design several times, so there is quite some code here related only to some Python versions

pyfakefs/mox3_stubout.py
```diff
@@ -102,7 +102,7 @@ def smart_set(self, obj, attr_name, new_attr):
         # function. We need to ensure that we put it back as a staticmethod.
         old_attribute = obj.__dict__.get(attr_name)
         if old_attribute is not None and isinstance(old_attribute, staticmethod):
-            orig_attr = staticmethod(orig_attr)
+            orig_attr = staticmethod(orig_attr)  # pytype: disable=not-callable
```
mrbean-bremen, line 105: could use a cast here, but that would not make it better...

pyfakefs/patched_packages.py
```diff
@@ -53,7 +53,7 @@ def get_modules_to_patch():
 def get_classes_to_patch():
     classes_to_patch = {}
     if patch_pandas:
-        classes_to_patch["TextFileReader"] = "pandas.io.parsers"
+        classes_to_patch["TextFileReader"] = ["pandas.io.parsers"]
```
mrbean-bremen, line 56: This was an actual bug

pyfakefs/fake_file.py
davidlbaird: Oh, egg on my face... it does not include None. I'm still not sure why path would return bytes. I see where AnyStr is useful for setting contents on a new file, but not why it is supported for anything else. Is it possible to implement the same annotations as implemented in the standard library?

davidlbaird: Again, answering my own question, I should really do some research before asking them, this annotation is close to the typeshed annotations. https://github.com/python/typeshed... but it might be a good idea to import them directly, if possible... for future investigation.

davidlbaird (approved):

pyfakefs/fake_file.py
mrbean-bremen: (this was in response to your previous comment, I hadn't seen your own response yet)

mrbean-bremen: I'm going to merge this as is now with the hope that the pytpe support can be improved with time.

PR #848, opened 2023-06-29
Add py.typed marker file (as defined by PEP 561)

jfgauron:
#### Describe the changes

Adds the `py.typed` file as per PEP 561. This ensures that projects that import pyfakefs can be type checked.

Note: First time working with setuptool so please make sure my change to `setup.cfg` is correct.

#### Tasks
- [x] ~~Unit tests added that reproduce the issue or prove feature is working~~ (not needed)
- [x] Fix or feature added
- [x] ~~Entry to release notes added~~ (not needed)
- [x] Unit tests passing
- [x] ~~For documentation changes: The Read the Docs preview builds and looks as expected~~ (not needed)

### DIFF
--- a/pyfakefs/py.typed
+++ b/pyfakefs/py.typed
@@ -1,0 +1,1 @@
+# Marker file for PEP 561.  The pyfakefs package uses inline types.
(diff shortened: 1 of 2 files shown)


mrbean-bremen (approved):
Thank you! I have waited with this as the typing is still not optimal in some places, but it is probably time to add this...

Issue #872, opened 2023-08-25
Is `fs` unique across tests?


jamesbraza:
I have something like the below code:

```python
import pathlib
import random
import time

import pytest
from pyfakefs.fake_filesystem import FakeFilesystem

FOLDER = (pathlib.Path(__file__).parent / "foo").resolve()
FOLDER.mkdir(exist_ok=True)
FILE = FOLDER / "file.csv"


def test_one(fs: FakeFilesystem) -> None:
    fs.create_file(FILE)
    time.sleep(random.random())
    assert FILE.exists()


@pytest.mark.usefixtures("fs")
def test_two() -> None:
    FILE.unlink(missing_ok=True)
    time.sleep(random.random() / 100)


def test_three(fs: FakeFilesystem) -> None:
    fs.create_file(FILE)
    FILE.unlink()
```

When running these tests via many processes in parallel using `tox run-parallel`, sometimes the test fails, which indicates to me the `fs` is not unique across test cases or test processes.

Is this the case?  I thought `fs` should be unique, but it seems like it isn't

jamesbraza:
This should be labelled as a "question" I think, sorry for clicking wrong button at first

jamesbraza:
Update: I think my issue is `create_file` doesn't actually patch in the file, you have to use the return of `create_file`.

So I needed to do something like this:

```python
def test_three(fs: FakeFilesystem) -> None:
    made_file = fs.create_file(FILE)
    pathlib.Path(made_file.path).unlink()
```

Feel free to close this as resolved, or leave open as an opportunity for better docs in `create_file`.

Also, one other suggestion is to make `FakeFile` objects subclasses of `pathlib.Path`.

mrbean-bremen:
> or leave open as an opportunity for better docs in create_file

I'll probaly add a link to the [API documentation](https://pytest-pyfakefs.readthedocs.io/en/latest/modules.html#pyfakefs.fake_filesystem.FakeFilesystem.create_file) in the [usage section](https://pytest-pyfakefs.readthedocs.io/en/latest/usage.html#file-creation-helpers), and adapt the summary of the API documentation make it more clear what it does.

> Also, one other suggestion is to make FakeFile objects subclasses of pathlib.Path.

As `FakeFile` is a mocked file (not a mocked `pathlib.Path` - that would be `FakePath`) and should be usable everywhere where a file can be used. As not every function handles `pathlib.Path` arguments, this is not feasible.

[...]

PR #873, opened 2023-08-25
Add some API links to the usage documentation

mrbean-bremen:
Also mention the return value of create_xxx methods more prominently.
Came up in #872

Added a note about global `pathlib.Path` objects to the troubleshooting section after the review discussion.

#### Tasks

### DIFF
--- a/docs/troubleshooting.rst
+++ b/docs/troubleshooting.rst
@@ -233,6 +233,30 @@ passed before the ``mocker`` fixture to ensure this:
       # works correctly
       mocker.patch("builtins.open", mocker.mock_open(read_data="content"))
 
+Pathlib.Path objects created outside of tests
+---------------------------------------------
+An pattern which is more often seen with the increased usage of `pathlib` is the
+creation of global `pathlib.Path` objects (instead of string paths) that are used
+in the tests. As these objects are created in the real filesystem, they are not of the same
+type as faked `pathlib.Path` objects, and both will always compare as not equal,
+regardless of the path they point to:
+
+.. code:: python
+
+  import pathlib
+
+  FOLDER = pathlib.Path(__file__).parent
+  FILE = FOLDER / "file.csv"
+
+
+  def test_one(fs):
+      fake_file = pathlib.Path(fs.create_file(FILE).path)
+      assert FILE == fake_file  # fails, compares different objects
+      assert str(FILE) == str(fake_file)  # succeeds, compares the actual paths
+
+Generally, mixing objects in the real filesystem and the fake filesystem
+is problematic and better avoided.
+
 
 .. _`multiprocessing`: https://docs.python.org/3/library/multiprocessing.html
 .. _`subprocess`: https://docs.python.org/3/library/subprocess.html
--- a/pyfakefs/fake_filesystem.py
+++ b/pyfakefs/fake_filesystem.py
@@ -183,7 +183,8 @@ class FakeFilesystem:
         is_windows_fs: `True` in a real or faked Windows file system.
         is_macos: `True` under MacOS, or if we are faking it.
         is_case_sensitive: `True` if a case-sensitive file system is assumed.
-        root: The root :py:class:`FakeDirectory` entry of the file system.
+        root: The root :py:class:`FakeDirectory<pyfakefs.fake_file.FakeDirectory>` entry
+            of the file system.
         umask: The umask used for newly created files, see `os.umask`.
         patcher: Holds the Patcher object if created from it. Allows access
             to the patcher object if using the pytest fs fixture.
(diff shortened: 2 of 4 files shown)


mrbean-bremen:
@jamesbraza - I made some minor changes to the documentation, please check if you are missing something.

pyfakefs/fake_filesystem.py
```diff
@@ -2034,7 +2034,8 @@ def make_string_path(self, path: AnyPath) -> AnyStr:
     def create_dir(
         self, directory_path: AnyPath, perm_bits: int = helpers.PERM_DEF
     ) -> FakeDirectory:
-        """Create `directory_path`, and all the parent directories.
+        """Create `directory_path` and all the parent directories, and return
+        the created FakeDirectory object.
```
jamesbraza, line 2038: Just wondering, why do you sometimes use `code` highlighting (`directory_path`) and sometimes not (FakeDirectory)?

jamesbraza (approved):
It may be good to add somewhere that the passed in `path` != `created_fake_file.path`

mrbean-bremen:
> It may be good to add somewhere that the passed in `path` != `created_fake_file.path`

Not quite sure what you mean. The created FakeFile should have the given path as its path (the string value of course in case you passed in a `Path` object). Can you elaborate?

pyfakefs/fake_filesystem.py
mrbean-bremen: > Just wondering, why do you sometimes use `code` highlighting (`directory_path`) and sometimes not (FakeDirectory)?

No reason. I actually was looking for some style guide as I wasn't sure if better to use single or double back ticks, but  got distracted and forgot about it..  Will have another look tomorrow.

jamesbraza:
> The created FakeFile should have the given path as its path (the string value of course in case you passed in a `Path` object). Can you elaborate?

I think the difference is whether the `pathlib.Path` has the `pathlib._PosixFlavour` or `pyfakefs.fake_pathlib._FakePosixFlavour`:

```python
import pathlib

from pyfakefs.fake_filesystem import FakeFilesystem

FOLDER = pathlib.Path(__file__).parent
FILE = FOLDER / "file.csv"


def test_one(fs: FakeFilesystem) -> None:
    fake_file = pathlib.Path(fs.create_file(FILE).path)
    assert FILE == fake_file  # This fails due to "flavour" differences
    # From what I have seen, using FILE from here can still use real fs,
    # while fake_file will only use the fake filesystem
    with FILE.open(mode="w", ...):
        ...
    with fake_file.open(mode="w", ...):
        ...
```

So I think perhaps maybe we should document `pathlib.Path`s made at the module-level, before `pyfakefs` can patch `pathlib`, may have some residual real filesystem usage.  What do you think?

mrbean-bremen:
> So I think perhaps maybe we should document `pathlib.Path`s made at the module-level, before `pyfakefs` can patch `pathlib`, may have some residual real filesystem usage. What do you think?

Ah ok - I thought this was about `create_file`, therefore I got confused.  
True, a fake `pathlib.Path` won't be equal to a real one (which is somewhat expected as they live in different file systems - in your example the real path is related to `__file__`, which does not exist in the fake file system). I cannot really change this - I could add an operator to the fake path for comparison with a real one, but not vice verse.
Maybe I could add a note in the troubleshouting guide, or in the limitations section...

pyfakefs/fake_filesystem.py
mrbean-bremen: Shall be fixed now.

docs/troubleshooting.rst
```diff
@@ -233,6 +233,30 @@ passed before the ``mocker`` fixture to ensure this:
       # works correctly
       mocker.patch("builtins.open", mocker.mock_open(read_data="content"))
 
+Pathlib.Path objects created outside of tests
```
jamesbraza, line 236: Thoughts on syntax highlighting as `pathlib.Path`?

docs/troubleshooting.rst
```diff
@@ -233,6 +233,30 @@ passed before the ``mocker`` fixture to ensure this:
...
+---------------------------------------------
+An pattern which is more often seen with the increased usage of `pathlib` is the
+creation of global `pathlib.Path` objects (instead of string paths) that are used
+in the tests. As these objects are created in the real filesystem, they are not of the same
```
jamesbraza, line 240:
```suggestion
creation of global `pathlib.Path` objects (instead of string paths) that are imported
into the tests. As these objects are created in the real filesystem, they are not of the same
```

docs/troubleshooting.rst
```diff
@@ -233,6 +233,30 @@ passed before the ``mocker`` fixture to ensure this:
...
+  import pathlib
+
+  FOLDER = pathlib.Path(__file__).parent
+  FILE = FOLDER / "file.csv"
+
+
+  def test_one(fs):
+      fake_file = pathlib.Path(fs.create_file(FILE).path)
+      assert FILE == fake_file  # fails, compares different objects
+      assert str(FILE) == str(fake_file)  # succeeds, compares the actual paths
```
jamesbraza, line 255:
```suggestion
  import pathlib

  # This Path was made in the real filesystem, before the test
  # stands up the fake filesystem
  FILE_PATH = pathlib.Path(__file__).parent / "file.csv"


  def test_path_equality(fs):
      # This Path was made after the fake filesystem is set up,
      # and thus patching within pathlib is in effect
      fake_file_path = pathlib.Path(fs.create_file(FILE_PATH).path)
      assert FILE_PATH == fake_file_path  # fails, compares different objects
      assert str(FILE_PATH) == str(fake_file_path)  # succeeds, compares the actual paths
```

Just cleaning up the snippet I provided:
- No need for `FOLDER`
- Removing room for confusion of `FILE` vague name:
    - `FILE` sounds like a file object when it's a `Path` object
    - Could detail if it's made in real or fake filesystem
- `test_one` name can be more specific

jamesbraza (approved):
Added a few comments, nice work! Mainly I am learning the "minimal repro" snippet can be even simpler, so I added some comments to streamline it

mrbean-bremen:
Thanks for the review - I'm going to merge this now. Hints to improve the documentation are always welcome!

PR #875, opened 2023-08-26
Removed extra create_file from troubleshooting snippet

jamesbraza:
- Noticed syntax highlighting needed an update
    - Single backtick means italics, double backtick means code syntax
- Fixed wording from "type" to "attributes"
- Realized to make the point in the code snippet about `Path`, don't need `create_file` at all

For https://github.com/pytest-dev/pyfakefs/issues/872

### DIFF
--- a/docs/troubleshooting.rst
+++ b/docs/troubleshooting.rst
@@ -235,10 +235,10 @@ passed before the ``mocker`` fixture to ensure this:
 
 Pathlib.Path objects created outside of tests
 ---------------------------------------------
-An pattern which is more often seen with the increased usage of `pathlib` is the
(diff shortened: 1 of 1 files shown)


mrbean-bremen:
Thanks - I will leave this open for a while, in case you find more stuff...

jamesbraza:
I already merged changes in my employer's codebase that implements all of these learnings, so the case is closed on my end.  Apologies for 2nd PR, always hunting for the true minimal repro 🔍 😄

mrbean-bremen:
No apologies needed 😀

mrbean-bremen (approved):

PR #1025, opened 2024-05-27
[WIP] Make fake filesystem, fake `os` and fake `os.path` respect `additional_skip_names`

sassanh:
#### Describe the changes
Add a call stack check for fake filesystem, `os` and `os.path` methods in their `handle_original_call` wrapper to see if they are being called from a module listed in `skip_modules`.

#### Tasks
- [x] Fix or feature added
- [x] Unit tests passing

### DIFF
--- a/CHANGES.md
+++ b/CHANGES.md
@@ -12,6 +12,12 @@ The released versions correspond to PyPI releases.
 * the default for `FakeFilesystem.shuffle_listdir_results` will change to `True` to reflect
   the real filesystem behavior
 
+## Unreleased
+
+### Enhancements
+
+- refactor the implementation of `additional_skip_names` parameter to make it work with more modules (see [#1023](../../issues/1023))
+
 ## [Version 5.6.0](https://pypi.python.org/pypi/pyfakefs/5.6.0) (2024-07-12)
 Adds preliminary Python 3.13 support.
 
--- a/pyfakefs/fake_os.py
+++ b/pyfakefs/fake_os.py
@@ -1421,9 +1421,13 @@ def handle_original_call(f: Callable) -> Callable:
 
         if not should_use_original and args:
             self = args[0]
+            fs: FakeFilesystem = self.filesystem
             if self.filesystem.patcher:
-                skip_names = self.filesystem.patcher._skip_names
-                if is_called_from_skipped_module(skip_names=skip_names):
+                skip_names = fs.patcher.skip_names
+                if is_called_from_skipped_module(
+                    skip_names=skip_names,
+                    case_sensitive=fs.is_case_sensitive,
+                ):
                     should_use_original = True
 
         if should_use_original:
--- a/pyfakefs/fake_os.py
+++ b/pyfakefs/fake_os.py
@@ -54,6 +54,7 @@ from pyfakefs.fake_path import FakePathModule
 from pyfakefs.fake_scandir import scandir, walk, ScanDirIter
 from pyfakefs.helpers import (
     FakeStatResult,
+    is_called_from_skipped_module,
     is_int_type,
     is_byte_string,
     make_string_path,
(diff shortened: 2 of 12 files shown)


sassanh:
This is mostly brainstorm material to see if we want to go in this direction.

Implementation details:
Most of the job is done as a filter in wrapper functions, `is_called_from_skipped_module` starts from the end of the call stack and goes up until it reaches a module that is:
1. Is not a frozen importlib
2. outside of Python's stdlib
3. outside of pyfakefs library itself (except for `tests/` and `pytest_tests/` directories)
This module is considered the module that is calling the wrapped function and if it is listed in `skip_names`, then it will try to run the original function instead of the fake wrapped function.

It is working for Python 3.12, 3.11 and mostly 3.10, but 3.9, 3.8 and 3.7 have yet different internals for pathlib.

sassanh:
@mrbean-bremen 
When you got the time, please check it and let me know if you approve this approach.
If it looks good to you, I will work on tests for other Python versions and Windows.

pyfakefs/helpers.py
```diff
@@ -426,3 +435,42 @@ def getvalue(self) -> bytes:
...
+            if not frame.filename.startswith("<frozen ")
+            and not frame.filename.startswith(STDLIB_PATH)
+            and (
+                not frame.filename.startswith(PYFAKEFS_PATH)
+                or any(
+                    frame.filename.startswith(test_path)
+                    for test_path in PYFAKEFS_TEST_PATHS
+                )
+            )
+        ),
```
mrbean-bremen, line 461: This looks like a good idea to me. I'm currently working on Python 3.13 compatibility, and the previous approach with the hardcoded stackframe size already does not work there, due to the `pathlib` module now implemented as a directory with the implementing modules inside (which also may call each other). It may decrease the performance a bit, but I hope that it will not be much.
I haven't used `sysconfig.get_path("stdlib")` before, but it looks like the correct way (it does handle global, user and virtual environments, right?).

pyfakefs/fake_os.py
```diff
@@ -1409,27 +1410,36 @@ def __getattr__(self, name: str) -> Any:
         return getattr(self.os_module, name)
 
 
-if sys.version_info > (3, 10):
```
mrbean-bremen, line 1412: This won't work with older Python versions, but you already know that...

mrbean-bremen (commented):
Thanks - as noted, I like that approach. There will be some work to do for the other Python versions (I don't think there is special handling needed for Windows), and we have to check the performance after the changes, but I think this is on a good way! 🚀

mrbean-bremen:
@sassanh - are you still working on this, or have you given up?

sassanh:
I'm not actively working on it but I have it in my radar.
If the open pull request is causing trouble in organization we can close it.
The blocking issue is still updating the tests for older python versions.

mrbean-bremen:
The PR isn't blocking anything - I just wanted to know if you are still on it, or if you need help. Take all the time you need - and thanks for the effort!

sassanh:
Feel free to add any commits or make any changes to this pull request as needed. Your help is greatly appreciated! I might not have the capacity to address tests for older versions in the new future.

mrbean-bremen:
@sassanh - I've rebased against main and fixed the failing tests (disabled one test for PyPy 3.7, as I didn't want to invest time in that outdated version), but other than that, only the release notes are missing as far as I can see.
Please check my commit and add the release notes (and anything you still want to add), and we can go and merge this.

sassanh:
@mrbean-bremen Thank you so much! I appreciate your help!
I will check your commit and add the release notes in the weekend.

sassanh:
@mrbean-bremen I checked the changes and it all looks good to me. I added the release notes too. Let's merge it if it looks good to you.

CHANGES.md
```diff
@@ -1,504 +1,600 @@
 # pyfakefs Release Notes
+
```
mrbean-bremen, line 2: These are a lot of changes - did you do some autoformatting here?

mrbean-bremen: Ok, I see that there are no real changes (e.g. the output won't change), so I will ignore this...

mrbean-bremen (approved):
Thanks for that! This change was something that I had in mind for quite some time, but never really got to it.

CHANGES.md
sassanh: I reverted the unrelated changes before seeing your second comment.

PR #1080, opened 2024-10-20
Add some support for loading fake modules

mrbean-bremen:
- works only if open_code patch mode is not off
 - see #1079

This will not handle all possible cases. Relative imports are not supported, and for absolute imports in dot notation an `__init__.py` is expected in module paths above the leaf module.

#### Tasks
- [x] Fix or feature added
- [x] Unit tests passing

### DIFF
--- a/CHANGES.md
+++ b/CHANGES.md
@@ -14,6 +14,10 @@ The released versions correspond to PyPI releases.
 
 ## Unreleased
 
+### Enhancements
+* added some support for loading fake modules in `AUTO` patch mode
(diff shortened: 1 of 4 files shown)


mrbean-bremen:
@Aran-Fey: I only added two tests (one of them basically your example test), please check if you need more cases covered.

Aran-Fey:
I'm surprised that relative imports don't work, but that's not a problem for me. This will help a lot, thanks!

mrbean-bremen:
> I'm surprised that relative imports don't work

You are right, this is easy to fix. Will do this.

PR #1144, opened 2025-03-15
Fix recursion on dynamic import from fake file

### LINKED ISSUE #1121: RecursionError if a fake module imports another fake module
**Describe the bug**
It's recently become possible to import modules from the fake file system, but if a fake module tries to import another fake module, it crashes with a RecursionError.

**How To Reproduce**
```python
import pytest, pathlib, sys
from pyfakefs.fake_filesystem import FakeFilesystem, PatchMode
from pyfakefs.fake_filesystem_unittest import Patcher


# Per default, importing from the fake file system doesn't work. So we must
# define a custom fixture which enables that.
@pytest.fixture
def fs():
    with Patcher(patch_open_code=PatchMode.AUTO) as p:
        yield p.fs


def test_dynamic_import(fs):
    pathlib.Path('foo.py').touch()
    pathlib.Path('bar.py').write_text('import foo')
    sys.path.insert(0, '')
    import bar


if __name__ == '__main__':
    pytest.main([__file__])
```

**Your environment**

Windows-11-10.0.22631-SP0
Python 3.13.1 (tags/v3.13.1:0671451, Dec  3 2024, 19:06:28) [MSC v.1942 64 bit (AMD64)]
pyfakefs 5.7.4
pytest 8.3.4

mrbean-bremen:
- happened in Python > 3.11 due to the handling of linecache in connection with a faked open_code call
- fixes #1121 

#### Tasks
- [x] Fix or feature added
- [x] Unit tests passing

### DIFF
--- a/pyfakefs/tests/fake_filesystem_unittest_test.py
+++ b/pyfakefs/tests/fake_filesystem_unittest_test.py
@@ -848,6 +848,13 @@ class AutoPatchOpenCodeTestCase(fake_filesystem_unittest.TestCase):
             self.import_foo()
         assert stdout.getvalue() == "hello\n"
 
+    def test_dynamic_import(self):
+        # regression test for #1121
+        self.fs.create_file("/foo.py")
+        self.fs.create_file("/bar.py", contents="import foo")
+        sys.path.insert(0, "")
+        import foo  # noqa
+
 
 class TestOtherFS(fake_filesystem_unittest.TestCase):
     def setUp(self):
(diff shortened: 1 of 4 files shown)


pyfakefs/tests/fake_filesystem_unittest_test.py
```diff
@@ -848,6 +848,12 @@ def test_exec_module_in_fake_fs(self):
             self.import_foo()
         assert stdout.getvalue() == "hello\n"
 
+    def test_dynamic_import(self):
+        # regression test for #1121
+        self.fs.create_file("/foo.py")
+        self.fs.create_file("/bar.py", contents="import foo")
+        sys.path.insert(0, "")
+
```
mrbean-bremen, line 856: Hm, ruff was too clever and has removed the import here...

Issue #1143, opened 2025-03-15
Including the fs fixture in another fixture requires a third fixture to be included in any test case using the fixture depending on fs


jfindlay:
**Describe the bug**
An unused fixture has to be included in a test case in order to prevent this error:
```
FileNotFoundError: [Errno 2] No such file or directory in the fake filesystem: '/tmp/pytest-of-jfindlay/pytest-92'
```
See the test case below for complete details.

**How To Reproduce**
* The important detail is at the bottom of `tests/test_case.py`.  One test case includes `mock_repo_root` and the other does not.  Neither test case uses the `mock_repo_root` fixture directly and so should not need to include it.

* The error does not occur if the `fs` fixture and associated is removed from the `mock_files` fixture.
### Test case
```python
# tests/test_case.py 
from pathlib import Path

import pytest
from pyfakefs.fake_filesystem import FakeFilesystem


@pytest.fixture()
def mock_repo_root(tmp_path: Path) -> Path:
    """
    Test repo root.
    """
    return tmp_path / "mock/repo/root"


@pytest.fixture()
def mock_files(
    # Error is introduced here ###
    fs: FakeFilesystem,
    ##############################
    mock_repo_root: Path
) -> list[Path]:
    """
    Mock some files and add them to the fake filesystem.
    """
    mocked_files = [
        mock_repo_root / "a/b/c",
        mock_repo_root / "a/b/d",
        mock_repo_root / "a/b/e",
    ]
    # `fs` fixture-related code ###
    for file in mocked_files:
        fs.create_file(file)
    ###############################
    return mocked_files


def test_files_works(mock_repo_root: Path, mock_files: list[Path]) -> None:
    assert mock_files


def test_files_errors(mock_files: list[Path]) -> None:
    assert mock_files
```
### Output
```console
$ pytest -vv -s -rA --tb=native tests/test_case.py
============================= test session starts ==============================
platform linux -- Python 3.11.2, pytest-7.2.0, pluggy-1.0.0+repack -- /usr/bin/python3
cachedir: .pytest_cache
tempdir: /tmp/saltfactories
rootdir: /etc/salt/repo/tests, configfile: pytest.ini
[...]

PR #1145, opened 2025-03-15
Add troubleshooting chapter for tmp_path caveat

mrbean-bremen:
- see #1143

#### Tasks

### DIFF
--- a/docs/troubleshooting.rst
+++ b/docs/troubleshooting.rst
@@ -266,6 +266,30 @@ passed before the ``mocker`` fixture to ensure this:
       # works correctly
       mocker.patch("builtins.open", mocker.mock_open(read_data="content"))
 
+tmp_path fixture with pyfakefs
+------------------------------
+If you are using the ``tmp_path`` fixture, or a similar pytest fixture
+relying on the real filesystem, you may have the opposite problem: now the ``fs``
+fixture must be added *after* the ``tmp_path`` fixture. Otherwise, the path will be
+created in the fake filesystem, and pytest will later try to access it in the real
+filesystem.
+
+.. code:: python
+
+  def test_working(tmp_path, fs):
+      fs.create_file(tmp_path / "foo")
+
+
+  def test_not_working(fs, tmp_path):
+      # causes an error while pytest tries to access the temporary directory
+      pass
+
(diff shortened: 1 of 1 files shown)


docs/troubleshooting.rst
```diff
@@ -266,6 +266,30 @@ passed before the ``mocker`` fixture to ensure this:
       # works correctly
       mocker.patch("builtins.open", mocker.mock_open(read_data="content"))
 
+tmp_path fixture with pyfakefs
```
jfindlay, line 269: I wonder if the wording here should be stronger:
> You do not need ``tmp_path`` if you are using ``pyfakefs``.  Both are methods designed to 1) protect the host filesystem and 2) guarantee a reproducible environment for each test case.  Both methods are completely adequate alone and combining them is unnecessary.  Tests that use both are likely to introduce unnecessary complexity and confusion.  For example, ..."

mrbean-bremen: Merged too soon...

Yes, thanks - I think you are right. I was wondering about it while I was writing it... There should be a general caution about fixtures relying on the real filesystem. I'll probably make a new PR tonight and this time wait for a review - or you could make one yourself if you want.

jfindlay: > Merged too soon...

Not necessarily.  I read through the full section and it makes more sense in context.  The idea formed in my mind many hours after reading the updated documentation and software is iterative and usually incomplete.  Thanks for the consideration.

> There should be a general caution about fixtures relying on the real filesystem

For me the main concerns cover two distinct cases:
1. Mixing `pyfakefs` with other testing affordances within the testing code that insulate the test case from the filesystem.
2. Writing code such that the unit test can completely mock all access points to files.

The second is much more complex and philosophical as, at least for python, this implies the user has to _actually write their code differently_ so that it can be tested (quite an imposition: complex or clever code requires awful contortions/tenuous introspections/egregious antipatterns just to cover the happy path with unit tests).  However, can pyfakefs give the user enough functionality such that they could reasonably cover any defensibly written code that is amenable to (unit) testing, for example, by sending in `FakePath` objects where the code expects `Path`?

I think the first case is simpler (though not completely orthogonal to the second) as it's easier to just proscribe `tmp_path` and rely on `fs` for filesystem abstractions in the test code itself.

PR #1163, opened 2025-04-16
Add missing `mode` property to fake file wrapper

### LINKED ISSUE #1162: Bug: FakeFileOpen objects don't implement the mode property required by typing.io.IO
**How To Reproduce**
```
from pyfakefs.fake_filesystem import FakeFilesystem
from pyfakefs.fake_open import FakeFileOpen
from pyfakefs.fake_pathlib import FakePathlibModule

fake_fs = FakeFilesystem()
fake_open = FakeFileOpen(fake_fs)
fake_pathlib = FakePathlibModule(fake_fs)

f = open("test.txt")
assert f.mode  # This works fine

fake_pathlib.Path("test.txt").write_text(f.read())
f_ = fake_open("test.txt")
assert f_.mode  # This fails
```

**Expected behavior:**
Fake file objects should properly implement the mode property as required by the typing.io.IO abstract class:

```
class IO(Generic[AnyStr]):
    @property
    @abstractmethod
    def mode(self) -> str:
        pass
```

**Your environment**
python3.9
pyfakefs==5.8.0

mrbean-bremen:
- conforms to the IO protocol
- fixes #1162

#### Tasks
- [x] Fix or feature added
- [x] Unit tests passing

### DIFF
--- a/pyfakefs/fake_file.py
+++ b/pyfakefs/fake_file.py
@@ -751,9 +751,8 @@ class FakeFileWrapper:
         self,
         file_object: FakeFile,
         file_path: AnyStr,
-        update: bool,
-        read: bool,
-        append: bool,
+        open_modes: _OpenModes,
+        allow_update: bool,
         delete_on_close: bool,
         filesystem: "FakeFilesystem",
         newline: Optional[str],
@@ -768,9 +767,8 @@ class FakeFileWrapper:
     ):
         self.file_object = file_object
         self.file_path = file_path  # type: ignore[var-annotated]
-        self._append = append
-        self._read = read
-        self.allow_update = update
+        self.open_modes = open_modes
+        self.allow_update = allow_update
         self._closefd = closefd
         self._file_epoch = file_object.epoch
         self.raw_io = raw_io
@@ -801,8 +799,8 @@ class FakeFileWrapper:
         self._flush_pos = 0
         if contents:
             self._flush_pos = len(contents)
-            if update:
-                if not append:
+            if self.allow_update:
+                if not self.open_modes.append:
                     self._io.seek(0)
                 else:
                     self._io.seek(self._flush_pos)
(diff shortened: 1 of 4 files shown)


pyfakefs/fake_file.py
```diff
@@ -890,6 +888,16 @@ def closed(self) -> bool:
         """Simulate the `closed` attribute on file."""
         return not self._is_open()
 
+    @property
+    def mode(self) -> str:
+        if self.open_modes.append:
```
evgenii-moriakhin, line 893: to clarify how I discovered this bug - it might help you better handle different values for the mode parameter:I came across this code in the imp module (imp.load_source) (Python 3.9):

```
class _HackedGetData:
    """Compatibility support for 'file' arguments of various load_*()
    functions."""

    def __init__(self, fullname, path, file=None):
        super().__init__(fullname, path)
        self.file = file

    def get_data(self, path):
        """Gross hack to contort loader to deal w/ load_*()'s bad API."""
        if self.file and path == self.path:
            # The contract of get_data() requires us to return bytes. Reopen the
            # file in binary mode if needed.
            if not self.file.closed:
                file = self.file
                if 'b' not in file.mode:
                    file.close()
            if self.file.closed:
                self.file = file = open(self.path, 'rb')

            with file:
                return file.read()
        else:
            return super().get_data(path)
```

This code relies on checking for "b" (binary) mode.
For completeness, it might make sense to add these values too

mrbean-bremen: Thanks, good point - I forgot that!

mrbean-bremen: Ok, added the check for binary mode, please check!

evgenii-moriakhin: from doc https://docs.python.org/3/library/functions.html#open

>> Modes 'r+' and 'r+b' open the file with no truncation.

in the current code here

```
       if self.open_modes.truncate:
            if self._binary:
                return "rb+" if self.open_modes.can_read else "wb"
```

this is probably a typo

although I see a separate test on r+b below, and maybe I haven't studied how the _truncate flag works in the fake file

mrbean-bremen: All these tests are run both in the real and in the fake fs, so no, this is not a typo - I just reproduced the behavior in the real file system, though that was unexpected to me, too.

mrbean-bremen: I also tested this locally, though I haven't found an explanation for this behavior.

mrbean-bremen: Ok, I'm merging this then.

mrbean-bremen: BTW, I've been thinking about the r+/w+ behavior, and I think that this is because while there is a difference on opening with r+ and w+ (w+ will truncate the file, r+ will not), _after_ the fie has been opened, the behavior is exactly the same for both modes (e.g. both allow reading and writing). Returning the mode in both cases as "r+" may be a design decision.

Issue #1171, opened 2025-05-23
Different ids for FakeOsModule.<some> Python3.12


evgenii-moriakhin:
On Python3.12 (python:3.12-alpine)

```python
import shutil
from unittest.mock import patch

from pyfakefs.fake_filesystem import FakeFilesystem
from pyfakefs.fake_os import FakeOsModule

fake_fs = FakeFilesystem()
os_mock = FakeOsModule(fake_fs)

for _ in range(10):
    print(id(os_mock.lstat))

with patch("shutil.os", os_mock):
    shutil.rmtree("/foo")

```

prints (sometimes) different ids and raised

```
Traceback (most recent call last):
  File "<stdin>", line 2, in <module>
  File "/usr/local/lib/python3.12/shutil.py", line 759, in rmtree
    _rmtree_safe_fd(stack, onexc)
  File "/usr/local/lib/python3.12/shutil.py", line 667, in _rmtree_safe_fd
    assert func is os.lstat
           ^^^^^^^^^^^^^^^^
AssertionError
```

I don't understand what this is related to, why different id's are returned for class

evgenii-moriakhin:
Yes

full output on python:3.12-alpine image for example

```
root@dynamix:~# docker run -it --rm python:3.12-alpine sh
/ # pip install pyfakefs
Collecting pyfakefs
  Downloading pyfakefs-5.8.0-py3-none-any.whl.metadata (7.8 kB)
Downloading pyfakefs-5.8.0-py3-none-any.whl (230 kB)
Installing collected packages: pyfakefs
Successfully installed pyfakefs-5.8.0
WARNING: Running pip as the 'root' user can result in broken permissions and conflicting behaviour with the system package manager, possibly rendering your system unusable. It is recommended to use a virtual environment instead: https://pip.pypa.io/warnings/venv. Use the --root-user-action option if you know what you are doing and want to suppress this warning.

[notice] A new release of pip is available: 25.0.1 -> 25.1.1
[notice] To update, run: pip install --upgrade pip
/ # python
Python 3.12.10 (main, May  9 2025, 23:41:15) [GCC 14.2.0] on linux
Type "help", "copyright", "credits" or "license" for more information.
>>> import shutil
>>> from unittest.mock import patch
>>> from pyfakefs.fake_filesystem import FakeFilesystem
>>> from pyfakefs.fake_os import FakeOsModule
>>> fake_fs = FakeFilesystem()
[...]

PR #1182, opened 2025-06-15
Move shutil patches to fake shutil module

mrbean-bremen:
- allows to use FakeShutilModule outside of Patcher
- see #1171

#### Tasks
- [ ] Unit tests added that reproduce the issue or prove feature is working - n/a
- [x] Fix or feature added
- [x] Unit tests passing

### DIFF
--- a/pyfakefs/fake_filesystem_shutil.py
+++ b/pyfakefs/fake_filesystem_shutil.py
@@ -22,11 +22,26 @@ work fine with the fake file system if `os`/`os.path` are patched.
     shutil module.
 
 :Usage:
+
+* With Patcher:
   The fake implementation is automatically involved if using
   `fake_filesystem_unittest.TestCase`, pytest fs fixture,
   or directly `Patcher`.
+
+* Stand-alone with FakeFilesystem:
+  To patch it independently of these, you also need to patch `os`, e.g:
+
+    filesystem = fake_filesystem.FakeFilesystem()
+    fake_os = fake_os.FakeOsModule(filesystem)
+    fake_shutil = fake_filesystem_shutil.FakeShutilModule(filesystem)
+
+    with patch("os", fake_os):
+        with patch("shutil.os", shutil_mock):
+            shutil.rmtree("path/in/fakefs")
+
 """
 
+import contextlib
 import os
 import shutil
 import sys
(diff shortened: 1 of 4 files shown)


pyfakefs/fake_filesystem_shutil.py
```diff
@@ -51,7 +86,46 @@ def __init__(self, filesystem):
...
+            if sys.version_info >= (3, 14):
+                self.__class__.shutil_module._rmtree_impl = (
+                    self.shutil_module._rmtree_safe_fd  # type: ignore[attr-defined]
+                )
+            else:
+                self.shutil_module._use_fd_functions = True
+
+    @contextlib.contextmanager
+    def patch_global_vars(self):
+        self.start_patching_global_vars()
```
evgenii-moriakhin, line 124: Shouldn't it be thread-safe? (adding locking) That is exactly the problem I described in the issue, which I solve not by patching but by injection. I also thought about such patching, but I didn't like the idea of locking, but since you made such a decision yourself in the code, maybe you should add it?

evgenii-moriakhin: I don't fully understand the code, though. Judging by what you're changing in patching, it's probably okay that multiple threads will do it.

pyfakefs/fake_filesystem_shutil.py
```diff
@@ -22,11 +22,26 @@
...
+
+* Stand-alone with FakeFilesystem:
+  To patch it independently of these, you also need to patch `os`, e.g:
+
+    filesystem = fake_filesystem.FakeFilesystem()
+    fake_os = fake_os.FakeOsModule(filesystem)
+    fake_shutil = fake_filesystem_shutil.FakeShutilModule(filesystem)
+
+    with patch("os", fake_os):
+        with patch("shutil.os", shutil_mock):
```
evgenii-moriakhin, line 39: And here

I still can't understand this example
What is shutil_mock, and why we replace shutil.>>os<< with it?

mrbean-bremen: Well, you can't understand it, because it is obviously a typo... 
I always see this stuff in code written by others, never in my own. 
Next time I will certainly wait for a review.

mrbean-bremen: Actually I think I'll just revert this documenation change, as there is currently there is no good way to support the standalone version. This has to wait for a better fix.

evgenii-moriakhin: Are you planning to revert only the usage example, not the new functionality itself? I just tested the library version from the main branch with my current product tests—no errors found

mrbean-bremen: Well, the problem is that I made an embarassing typo ("start" instead of "stop") that led to the patches not to be reverted. This may not be a problem if you only want to use the patched version as in your case, but is if afterwards the real code should be executed. It would not be noticable either, because the patches just change the code path so that some optimized variants are not used that cannot easily by patched, but this makes potential subtle problems harder to detect.

Anyway, I cannot release the code as is, and after fixing the typo, the tests do not pass in the CI under macOS. Unfortunately, I cannot test this locally, so it may take a while to resolve this. Also, I'm short on time this week (and possibly the next), and I also need a bit more sleep to avoid such dumb mistakes.

TL;DR - it is unlikely that I will finish this soon, certainly not before the next weekend.

PR #1185, opened 2025-06-20
Add thread lock for global patches

mrbean-bremen:
- remove incorrect documentation about stand-alone usage
- fix patching of global variables

This mainly fixes the latest PR that was quite buggy, and had added incorrect documentation, see #1171 and #1182.

#### Tasks
- [ ] Unit tests added that reproduce the issue or prove feature is working - n/a
- [x] Fix or feature added
- [x] Unit tests passing

### DIFF
--- a/pyfakefs/fake_filesystem_shutil.py
+++ b/pyfakefs/fake_filesystem_shutil.py
@@ -23,28 +23,17 @@ work fine with the fake file system if `os`/`os.path` are patched.
 
 :Usage:
 
-* With Patcher:
-  The fake implementation is automatically involved if using
-  `fake_filesystem_unittest.TestCase`, pytest fs fixture,
-  or directly `Patcher`.
-
-* Stand-alone with FakeFilesystem:
-  To patch it independently of these, you also need to patch `os`, e.g:
-
-    filesystem = fake_filesystem.FakeFilesystem()
-    fake_os = fake_os.FakeOsModule(filesystem)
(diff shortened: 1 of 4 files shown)


pyfakefs/fake_filesystem_shutil.py
```diff
@@ -113,19 +95,28 @@ def stop_patching_global_vars(self):
...
-        finally:
-            self.start_patching_global_vars()
+    def with_patched_globals(self, f: Callable) -> Callable:
+        """Function wrapper that patches global variables during function execution.
+        Can be used in multi-threading code.
+        """
+
+        @functools.wraps(f)
+        def wrapped(*args, **kwargs):
+            lock = Lock()
```
evgenii-moriakhin, line 111: Is it correct to create a different instances of Lock here on each call? If we are trying to protect `self.__class__` global variables from modification (and conflicts) in multithreaded code, shouldn't there be one instance of Lock for the whole class? `with self.__class__._patch_globals_lock: ...`

mrbean-bremen: Hadn't seen your reply - if you don't mind, I would leave it as is.

mrbean-bremen: The private global variables define if some optimized path shall be taken (which usually involves some OS-specifc calls that can't be patched easily).

evgenii-moriakhin: > Hadn't seen your reply - if you don't mind, I would leave it as is.

Of course, it's enough that we've discussed it

PR #1212, opened 2025-08-19
Fix the pinned build-system version

kurtmckee:
#### Describe the changes

setuptools v45.x doesn't support `pyproject.toml` files.
v62.1.x and higher do, and continue to support Python 3.7.

Reproducer:

```
# Create a venv
python -m venv venv-demo
source venv-demo/bin/activate

# Install setuptools and build
python -m pip install setuptools==45.0.0 build

# Build the package using the already-installed setuptools
python -m build --no-isolation
```

This results in the following error:

```
ERROR Backend 'setuptools.build_meta' is not available.
```

Changing the reproducer to use setuptools 61.2.0 demonstrates that the issue is resolved.

#### Tasks
- [ ] Fix or feature added
- [x] Unit tests passing

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -1,12 +1,13 @@
 [build-system]
-requires = ["setuptools>=45"]
+# change this to be "setuptools>=77.0.3" after 3.8 is out of support
+requires = ["setuptools>=61.2"]
 build-backend = "setuptools.build_meta"
 
 [project]
 name = "pyfakefs"
 description = "Implements a fake file system that mocks the Python file system modules."
 readme = "README.md"
-# change this to be PEP-639-conform after 3.8 is out of support
+# change this to be "Apache-2.0" after 3.8 is out of support
 license = {file = "COPYING"}
 
 authors = [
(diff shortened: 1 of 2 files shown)


mrbean-bremen:
Thanks, this looks sensible, as far as I can see! 

I'm traveling right now, will have a closer look / merge next week when I have access to a computer.

pyproject.toml
```diff
@@ -1,12 +1,13 @@
...
+# change this to be "setuptools>=77.0.3" after 3.8 is out of support
+requires = ["setuptools>=61.2"]
 build-backend = "setuptools.build_meta"
 
 [project]
 name = "pyfakefs"
 description = "Implements a fake file system that mocks the Python file system modules."
 readme = "README.md"
-# change this to be PEP-639-conform after 3.8 is out of support
+# change this to be "Apache-2.0" after 3.8 is out of support
```
mrbean-bremen, line 10: Actually I would leave the old comment. PEP-639-conform means that we remove the license from the classifiers and move it here _and_ add the license file (in `license-files`).

mrbean-bremen (approved):
Thanks!

PR #1318, opened 2026-04-23
Make use_original thread-local to fix cross-thread races

### LINKED ISSUE #1317: `FakeOsModule.use_original` is not thread-safe under `asyncio.to_thread` workloads
_[This bug was found, isolated, fixed, and reported by Claude Opus 4.7, with some minor edits by me.]_

**Describe the bug**

`FakeOsModule.use_original` is a process-global class attribute flipped by the `use_original_os()` context manager. When one thread enters the context manager while another thread is mid-faked-filesystem-operation, the second thread observes `use_original=True` set by the first thread and dispatches to the real `os` module, raising a real `PermissionError` against paths that only exist in the fake filesystem.

_Expected_: every faked filesystem call inside `Patcher` stays faked, regardless of what other threads are doing.

_Observed_: intermittent `PermissionError: [Errno 13] Permission denied: '<fake-path>'` from real `os.mkdir` while a worker thread is calling `Path.mkdir()` against the fake filesystem.

Representative stack trace (from the plain repro below on 6.2.0):
```python
  File ".../threading.py", line 1073, in _bootstrap_inner
    self.run()
  File ".../threading.py", line 1010, in run
    self._target(*self._args, **self._kwargs)
  File ".../concurrent/futures/thread.py", line 59, in run
    result = self.fn(*self.args, **self.kwargs)
  File ".../repro.py", line 6, in _write
    p.parent.mkdir(parents=True, exist_ok=True)
  File ".../pathlib/_local.py", line 770, in mkdir
    os.mkdir(self, mode)
PermissionError: [Errno 13] Permission denied: '/fake-env'
```
The trace shows `pathlib.Path.mkdir` → `os.mkdir(self, mode)` reaching the real kernel's `mkdir`, which has no knowledge of `/fake-env` (the path exists only inside pyfakefs's in-memory tree).

**How To Reproduce**

_Plain repro_:
```python
import asyncio
from pathlib import Path
from pyfakefs.fake_filesystem_unittest import Patcher

def _write(p):
    p.parent.mkdir(parents=True, exist_ok=True)
    with p.open("w") as cf:
        cf.write("x")

async def _dispatch(p):
    await asyncio.to_thread(_write, p)

async def many():
    await asyncio.gather(*[
        _dispatch(Path(f"/fake-env/a/m{i}/deep/file")) for i in range(8)
    ])

with Patcher() as p:
    p.fs.create_dir("/fake-env")
    asyncio.run(many())
```
The repro's failure rate depends on whether another pyfakefs-internal code path (linecache during traceback formatting, `RealPath` wrappers in `fake_pathlib`, etc.) concurrently enters `use_original_os()` while the workers run.

_Deterministic repro_: fail reliably by adding hammer threads entering the context manager in a tight loop from within the `Patcher` context so they do not contend with pyfakefs's cold init:

```python
import asyncio
import threading
from pathlib import Path
from pyfakefs.fake_filesystem_unittest import Patcher
from pyfakefs.fake_os import use_original_os

def _write(p):
    p.parent.mkdir(parents=True, exist_ok=True)
    with p.open("w") as cf:
        cf.write("x")

async def _dispatch(p):
    await asyncio.to_thread(_write, p)

async def many(root, n):
    await asyncio.gather(
        *[_dispatch(Path(f"{root}/a/m{i}/deep/file")) for i in range(n)]
    )

with Patcher() as p:
    stop = threading.Event()
    def hammer():
        while not stop.is_set():
            with use_original_os():
                pass
    hammers = [threading.Thread(target=hammer, daemon=True) for _ in range(4)]
    for h in hammers:
        h.start()
    try:
        for round_idx in range(3):
            root = f"/fake-env-{round_idx}"
            p.fs.create_dir(root)
            asyncio.run(many(root, 32))
    finally:
        stop.set()
        for h in hammers:
            h.join(timeout=1)
```
With hammers running inside the `Patcher` context, the failure is reliable.

**Your environment**
```
Linux-6.17.0-20-generic-x86_64-with-glibc2.39
Python 3.12.3 (main, Mar  3 2026, 12:15:18) [GCC 13.3.0]
pyfakefs 6.2.0
pytest 9.0.2
```
Also reproduces on current `main` (6.3.dev0, commit `b907b2b`).

jfindlay:
_[This bug was found, isolated, fixed, and reported by Claude Opus 4.7, with some minor edits by me.  You are welcome to update or adapt as needed.]_

Fixes #1317.

#### Describe the changes

The `use_original` flag on `FakeOsModule` was a process-global class attribute toggled by the `use_original_os()` context manager. When one thread entered the context manager while another thread was in the middle of a faked filesystem operation, the faked call could observe `use_original=True` and dispatch to the real `os` module, failing against paths that only exist in the fake filesystem.

Store `use_original` in a `threading.local` and have `use_original_os()` save and restore the previous thread-local value so nested uses continue to work.  The class attribute becomes an instance property, preserving the existing `instance.use_original` get/set API.

A regression test in `fake_filesystem_unittest_test` exercises the concurrent-worker pattern with explicit hammer threads entering `use_original_os()` alongside `asyncio.to_thread` workers; without the fix it fails reliably with `PermissionError` against the fake root.

#### Tasks
- [x] Fix or feature added
- [x] Unit tests passing
- [-] For documentation changes: The Read the Docs preview builds and looks as expected

### DIFF
--- a/CHANGES.md
+++ b/CHANGES.md
@@ -7,6 +7,9 @@ The released versions correspond to PyPI releases.
 ### Fixes
 * fixed a crash if the stack limit was set to a low value
   (see [#1313](https://github.com/pytest-dev/pyfakefs/issues/1313))
+* made the `use_original` flag on `FakeOsModule` thread-local to fix a race
+  condition with async code (see
+  [#1317](https://github.com/pytest-dev/pyfakefs/issues/1317))
 
 ## [Version 6.2.0](https://pypi.python.org/pypi/pyfakefs/6.2.0) (2026-04-12)
 Changes the MRO for file wrappers.
--- a/pyfakefs/fake_os.py
+++ b/pyfakefs/fake_os.py
@@ -23,6 +23,7 @@ import functools
 import inspect
 import os
 import sys
+import threading
 import uuid
 from contextlib import contextmanager
 from stat import (
@@ -74,6 +75,15 @@ if TYPE_CHECKING:
 NR_STD_STREAMS = 3
 
 
+# Thread-local storage for the `use_original` flag.
+_thread_state = threading.local()
+
+
+def _use_original() -> bool:
+    """Return the thread-local `use_original` flag, defaulting to False."""
+    return getattr(_thread_state, "use_original", False)
+
+
 class FakeOsModule:
     """Uses FakeFilesystem to provide a fake os module replacement.
 
@@ -88,7 +98,18 @@ class FakeOsModule:
         my_os_module = fake_os.FakeOsModule(filesystem)
     """
 
-    use_original = False
+    @property
+    def use_original(self) -> bool:
+        """Whether faked operations should dispatch to the real `os` module.
+
+        Stored in thread-local state so that `use_original_os()` in one thread
+        does not affect faked operations running in another thread.
+        """
+        return _use_original()
+
+    @use_original.setter
+    def use_original(self, value: bool) -> None:
+        _thread_state.use_original = value
 
     @staticmethod
     def dir() -> list[str]:
@@ -1468,7 +1489,7 @@ def handle_original_call(f: Callable) -> Callable:
 
     @functools.wraps(f)
     def wrapped(*args, **kwargs):
-        should_use_original = FakeOsModule.use_original
+        should_use_original = _use_original()
 
         if not should_use_original and args:
             self = args[0]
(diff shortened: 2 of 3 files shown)


mrbean-bremen:
"Unit tests passing" is not completely true - the new test is failing in one build. I haven't looked into it yet, may have a closer look tonight...

mrbean-bremen:
Ok, I can reproduce the test failure locally under WSL/ubuntu with pypy 3.10, and it looks like this is another threading issue related to the use of `heapq`. Probably one of the things you had already found...

PR #1322, opened 2026-05-26
AI-generated possible solution for #1317 (A)

hugnia:
Hi,

We're a research team at JetBrains studying how maintainers evaluate AI-generated code.

We opened two pull requests as possible solutions to the corresponding issue, each taking a different approach. Both were generated by AI and have not yet been manually reviewed by a human. We’d like to get your perspective on these two solutions via a 10-minute survey. Regardless of your role, we want to know which solution actually saves you work, and which qualities of AI-generated code matter most when working with the codebase.

If you're open to participating in our study, just reply "I'm interested" here, and we'll send the details to the contact specified in your GitHub profile. If you'd rather not take part, feel free to close the PRs – we won’t post them again.

Thanks!

Issue #1317

### DIFF
--- a/pyfakefs/fake_os.py
+++ b/pyfakefs/fake_os.py
@@ -87,6 +87,7 @@ def _set_use_original(value: bool) -> None:
     """Set the current thread use_original flag value."""
     _use_original_local.value = value
 
+
 class FakeOsModule:
     """Uses FakeFilesystem to provide a fake os module replacement.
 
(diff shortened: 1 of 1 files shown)


mrbean-bremen:
Thanks - I will have a closer look at this after I'm back from vacation.

mrbean-bremen:
Actually - no need to wait. This issue has already a PR with a solution (also mostly AI-generated) by another contributor, which tries to solve the general issue of thread-safety instead of the specific problem. I will have a closer look at that solution after I'm back from vacation, and probably won't consider this PR, as it is superseded by the other one.

hugnia:
Thank you for taking the time to look into this. Wishing you a great vacation!

PR #1342, opened 2026-09-22
Fix os.rename permission check for the root user

### LINKED ISSUE #1341: os.rename fails with `PermissionError` when the parent directory belongs to another user
**Describe the bug**

For the root user, `os.rename()` fails with `PermissionError` when the parent directory belongs to another user. `open()` and `os.stat()` on the same file still work. On a real filesystem, real root can rename the file. So pyfakefs is stricter than the real filesystem, but only for uid 0.

**How To Reproduce**

```python
import os
from pyfakefs import fake_filesystem, fake_filesystem_unittest


class TestRenameAsRoot(fake_filesystem_unittest.TestCase):
    def setUp(self):
        self.setUpPyfakefs()
        fake_filesystem.set_uid(0)

    def test_rename_as_root(self):
        self.fs.create_dir("/d", perm_bits=0o700)
        os.chown("/d", 8765, 8765)
        open("/d/tmp", "w").close()  # ok
        os.stat("/d/tmp")            # ok
        os.rename("/d/tmp", "/d/x")  # PermissionError: [Errno 13] ... '/d'
```

**Expected behaviour**

I ran the same three calls on a real tmpfs, as real uid 0, with the directory set to mode 0700 and owned by 8765:8765:

```
uid=0  dir mode=0o700 owner=8765:8765
  open  : ok
  stat  : ok
  rename: ok
```

**Your environment**

```
Linux-7.0.14-16-pve-x86_64-with-glibc2.41
Python 3.13.5 (main, Aug 10 2026, 12:06:59) [GCC 14.2.0]
pyfakefs 6.2.0
pytest 9.1.1
```

I also reproduced it on current main, commit c72885a:

```
Linux-7.0.14-16-pve-x86_64-with-glibc2.41
Python 3.13.5 (main, Aug 10 2026, 12:06:59) [GCC 14.2.0]
pyfakefs 6.3.dev0
pytest 9.1.1
```

**When it started**

In 5.4.0, with #961, which fixed #959 and #960. That change replaced direct `st_mode` tests with the new `has_permission()`. Where an `is_root()` check was already there, the change kept it. But in `lresolve()` the old line was:

```python
if not parent_obj.st_mode & helpers.PERM_READ:
```

This line did not need an `is_root()` check, because `0o700 & 0o400` is true no matter who owns the directory. The new call also looks at the owner, so now the missing check matters.

The test added for #959, `test_fail_add_entry_to_readonly_dir`, skips when it runs as root. So this case was never tested.

tchaikov:
- lresolve() was the only permission check without an is_root() test, so root could not rename an entry in a directory of another user
- regression from #961, where the plain mode check was replaced by has_permission(), which also considers the owner
- fixes #1341

#### Describe the changes
The related issue or a description of the bug or feature that this PR addresses.

#### Tasks
- [x] Fix or feature added
- [x] Unit tests passing


mrbean-bremen (approved):
Thanks, looks good!
The failing tests are another problem, some flaky tests with the new macOS version...

